PR #36, opened 2018-02-28
added travis ci config file

matthewshim-ms:
See [Travis-CI Docs](https://docs.travis-ci.com/user/languages/python/), Repo owner is going to have to manage Travis CI permissions

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,5 @@
+language: python
+python:
+  - "3.6"
+install:
+  - pip install -r requirements.txt


.travis.yml
```diff
@@ -0,0 +1,5 @@
+language: python
+python:
+  - "3.6"
+install:
+  - pip install -r requirements.txt
```
johnataylor: Did you test this?
Do we need to update the requirements.txt?

matthewshim-ms: The best way to test this is configure the Repo on the Travis portal, then make any push and it should "build" - currently there is no requirements.txt so we may need to add this. Configuration should be easy - see image 

image

johnataylor: Sure so you are saying "no" and "yes" OK let's do that then. We don't need to take this PR until we have those then.

matthewshim-ms: I just tested locally and integration with Travis-CI worked, just need requirements.txt and the Repo owner needs to configure it in the Travis portal

stevengum97:
What's going on between #36 and #38? Which one are we keeping as the one we're planning on merging?

stevengum97:
Closing this PR as #38 was merged yesterday. If we still need this one for some reason we can reopen it.

PR #43, opened 2018-03-16
Aiohttp sample

stevengum97:


libraries/botbuilder-core/botbuilder/core/middleware_set.py
```diff
@@ -0,0 +1,58 @@
...
+    async def on_process_request(self, context, logic: Callable):
+        await self.run(context, logic)
+
+    async def run(self, context: BotContext, logic: Callable):
+        handlers = self._middleware[0:]
+
+        async def run_next(i: int):
+            try:
+                if i < len(handlers):
+                    await run_next(handlers[i](context, run_next(i+1)))
```
johnataylor: this might read better without the recursion. presumably teh recursion came from the JavaScript Promises

PR #81, opened 2018-05-29
add CardFactory, MessageFactory and tests

stevengum97:
- Add `CardFactory` and its tests
- Add `MessageFactory` and its tests

### DIFF
--- a/libraries/botbuilder-core/botbuilder/core/message_factory.py
+++ b/libraries/botbuilder-core/botbuilder/core/message_factory.py
@@ -3,9 +3,7 @@
 
 from typing import List, Union
 from botbuilder.schema import (ActivityTypes, Activity, Attachment,
-                               AttachmentLayoutTypes, CardAction, SuggestedActions, InputHints)
-
-from .card_factory import CardFactory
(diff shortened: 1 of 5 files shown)


libraries/botbuilder-core/botbuilder/core/message_factory.py
```diff
@@ -0,0 +1,166 @@
...
+        :param text:
+        :param speak:
+        :param input_hint:
+        :return:
+        """
+        return attachment_activity(AttachmentLayoutTypes.carousel, attachments, text, speak, input_hint)
+
+    @staticmethod
+    def content_url(url: str, content_type: str, name: str = None, text: str = None, speak: str = None,
+                    input_hint: Union[InputHints, str] = None):
```
johnataylor: so we are still keeping this one overload?
- I'm thinking of simplifying the C# to just take the object in cases like this.
- but we can check in with this and resolve that later

stevengum97: Actually we need to keep that to account for the _not-Swagger-defined_ `"delay"`-type activity. Without this overload we'd be saying no to this specific activity type.

PR #84, opened 2018-06-16
Johtaylo/fix bld

johnataylor:


johnataylor:
This doesn't fix the problems... quick summary of where we're at with this...
We have in the past checked in generated code - we think that is incompatible with the latest rev of the autorest runtime. We are investigating. We wanted to use the later revs of the autorest code only because it got us the async build of the client stubs. We are re-evaluating all this.

PR #109, opened 2018-10-03
Add new sample bot

SynneK1337:
I created simple, but quite useful bot for every Giżycko Citizien. It allows you to check the weather and check is the rotational bridge is open or not.


samples/Console-GizyckoBot/main.py
```diff
@@ -0,0 +1,83 @@
...
+        return True
+    else:
+        return False
+
+
+class Weather():
+    http = urllib3.PoolManager()
+
+    def get_weather(self):
+        r = self.http.request('GET', 'http://api.openweathermap.org/data/2.5/weather?q=Gizycko,pl&APPID=0ef268648751f483a62fd90feb04c5eb&units=metric')
```
stevengum: I recommend using an async library as this call is blocking.

samples/Console-GizyckoBot/main.py
```diff
@@ -0,0 +1,83 @@
...
+        r = json.loads(r.data.decode('utf-8'))
+        self.status = r["weather"][0]["description"]
+        self.temp = int(r["main"]["temp"])
+        self.wind_speed = r["wind"]["speed"]
```
stevengum: You're storing the retrieved data on the class itself, but without further usage of this data, it would be better to return the resulting formatted data. Looking at your code it creates a new `Weather` instance on every turn, so the data is never reused.

Additionally, it's generally frowned (in multiple languages) upon to create class properties or methods without declaring them first in the initializer or constructor.

stevengum:
Thank you for the contribution, this is awesome! 👍 

 Unfortunately I'm going to close this PR as it's out of scope for the repository. We are not at this time taking location-specific samples.

I recommend creating a new repository and moving your sample (and potentially other BB-Py samples you may do) there. Thank you again for the PR!

PR #110, opened 2018-10-04
Azure CosmosDbStorage Middleware

eavanvalkenburg:
Hi all, 

I created a middleware package for talking with CosmosDB from the Python Bot Framework, I used a similar structure with a botbuilder-azure package. It includes a config object and a storage object, as well as tests for the integration. In the samples folder is a adaptation of the EchoBot that talks with Cosmos instead of memory storage. Automated testing will rely on CosmosDB emulator running, this is the same as with the Node version of the CosmosDB storage.

## Most important pieces:

- botbuilder-azure/botbuilder/azure/cosmosdb_storage.py - main code for using Cosmos as a middleware
- botbuilder-azure/tests/test_cosmos_storage.py - tests for the middleware
- samples/Echobot-with-CosmosState/* - sample that uses the middleware

### DIFF
--- a/libraries/botbuilder-azure/botbuilder/azure/about.py
+++ b/libraries/botbuilder-azure/botbuilder/azure/about.py
@@ -2,7 +2,7 @@
 # Licensed under the MIT License.
 
 __title__ = 'botbuilder-azure'
-__version__ = '0.1'
+__version__ = '4.0.0.a6'
 __uri__ = 'https://www.github.com/Microsoft/botbuilder-python'
 __author__ = 'Microsoft'
 __description__ = 'Microsoft Bot Framework Bot Builder'
--- a/libraries/botbuilder-azure/botbuilder/azure/about.py
+++ b/libraries/botbuilder-azure/botbuilder/azure/about.py
@@ -1,0 +1,10 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+__title__ = 'botbuilder-azure'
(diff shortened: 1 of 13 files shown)


eavanvalkenburg:
I disabled the tests that talk to CosmosDB, similar to the JS version of this package since there is no CosmosDB emulator for Linux, tests run locally against a online Cosmos work. Alternative would be to run these tests on Windows since the emulator works there.

libraries/botbuilder-azure/botbuilder/azure/about.py
```diff
@@ -0,0 +1,10 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+__title__ = 'botbuilder-azure'
+__version__ = '0.1'
```
stevengum: Instead of `'0.1'`, please use the latest preview version: `'4.0.0.a6'`.

libraries/botbuilder-azure/setup.py
```diff
@@ -0,0 +1,40 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+from setuptools import setup
+import os
+
+NAME = "botbuilder-azure"
+VERSION = "0.1"
```
stevengum: This isn't used and can be removed.

eavanvalkenburg:
Hey @stevengum any updates on this? I would love to have this approved in October so that it counts towards my #hacktoberfest progress!

PR #132, opened 2019-03-16
Fix Markdown formatting in samples README files

v-kydela:


v-kydela:
I don't know why this received no attention for two months, but I'm closing it

PR #148, opened 2019-04-26
Feature parity with C# and JS

Zerryth:
* QnAMaker class with basic features implemented

**working on:**
* telemetry
* tests
* nice things like throwing errors


stevengum:
@daveta, @congysu, @axelsrz, @Zerryth Why is everyone working on the same branch and touching different libraries?

Reviewers who like to use Code Flow cannot use the tool to review this PR due to the number of commits. Please consider getting this PR in ASAP and then quickly iterating with _separate_ PRs. Otherwise this PR is going to be close to impossible to review.

Zerryth:
@stevengum I have a sneaking suspicion the reason is "b/c Ashley messed up with git/github at some point >_>"

axelsrz:
@stevengum @Zerryth Actually what happened is that I was using this branch 3 weeks ago to push the general template for the new classes and my guess is that because a lot of breaking code was pushed in a small amount of time, this branch was used as quick push. We should start to at least make PRs against this branch (work-in-progress). I'm almost done with building the fix so we can do that from now on.

PR #149, opened 2019-04-30
Fix waterfall test (add accessor set)

daveta:
Fix waterfall test.  

The DialogState was not being set() (via accessor) before the storage save occurred.

### DIFF
--- a/libraries/botbuilder-applicationinsights/tests/test_telemetry_waterfall.py
+++ b/libraries/botbuilder-applicationinsights/tests/test_telemetry_waterfall.py
@@ -23,6 +23,7 @@ from botbuilder.core import (
 from botbuilder.dialogs import (
                                 Dialog,
                                 DialogSet,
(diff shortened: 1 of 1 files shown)


libraries/botbuilder-applicationinsights/tests/test_telemetry_waterfall.py
```diff
@@ -45,7 +46,7 @@ def test_none_telemetry_client(self):
         # assert
         self.assertEqual(type(dialog.telemetry_client), NullTelemetryClient)
 
-    @skip('Pending Telemetry mock')
+
     @patch('test_telemetry_waterfall.ApplicationInsightsTelemetryClient')
```
axelsrz: could you change this (and line 107) to: `@patch('botbuilder.applicationinsights.ApplicationInsightsTelemetryClient')` for the build to pass?

PR #153, opened 2019-05-01
serialization: py json interop

congysu:
* serialize for RecognizerResult, LuisResult etc.
* add test for telemetry
* add test for trace activity

### DIFF
--- a/libraries/botbuilder-ai/botbuilder/ai/luis/luis_prediction_options.py
+++ b/libraries/botbuilder-ai/botbuilder/ai/luis/luis_prediction_options.py
@@ -11,18 +11,25 @@ class LuisPredictionOptions(object):
 
     def __init__(
         self,
+        bing_spell_check_subscription_key: str = None,
+        include_all_intents: bool = None,
+        include_instance_data: bool = None,
+        log: bool = None,
+        spell_check: bool = None,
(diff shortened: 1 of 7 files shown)


libraries/botbuilder-ai/botbuilder/ai/luis/luis_prediction_options.py
```diff
@@ -9,7 +9,12 @@ class LuisPredictionOptions(object):
     Optional parameters for a LUIS prediction request.
     """
 
-    def __init__(self, timeout: float = 100000):
+    def __init__(
+        self,
+        timeout: float = 100000,
+        telemetry_client: BotTelemetryClient = NullTelemetryClient,
+        log_personal_information: bool = False,
+    ):
```
daveta: How "to the letter of the law" are we going to be following C#?  I like this as a usability feature.
[https://github.com/Microsoft/botbuilder-dotnet/blob/master/libraries/Microsoft.Bot.Builder.AI.LUIS/LuisPredictionOptions.cs](https://github.com/Microsoft/botbuilder-dotnet/blob/master/libraries/Microsoft.Bot.Builder.AI.LUIS/LuisPredictionOptions.cs)

PR #165, opened 2019-05-07
remove Travis CI badge from README.md

stevengum:
The link doesn't work (because of the org "rename") but also the last build against Travis CI was over 2 months ago; and Azure Dev Ops is setup.


stevengum:
Done!

PR #176, opened 2019-05-14
Flask Core-Bot Sample

daveta:
Adopted the naming style for when we merge into botbuilder-samples:
- python-flask
   - 13.core-bot

Same sample as always, salient things:
- For configuration, adopted simplest flask-supported file-based configuration.  
http://flask.pocoo.org/docs/1.0/config/
- Using asyncio Tasks.
- Initially turned on debug.

I'll convert the App Insights sample as well later.


samples/python-flask/13.core-bot/main.py
```diff
@@ -0,0 +1,59 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+"""
+This sample shows how to create a simple EchoBot with state.
```
axelsrz: Not EchoBot, we should change this in the other sample as well.

daveta: I'll fix this on the next checkin - coming shortly..

Issue #167, opened 2019-05-07
Migrate onTurnError functionality to BotAdapter


axelsrz:
I didn't see an `adapter.onTurnError()` equivalent in the sample, is this in the library's implementation?

_Originally posted by @stevengum in https://github.com/microsoft/botbuilder-python/pull/164_

axelsrz:
Solved in #180

PR #180, opened 2019-05-20
On turn error on Bot Adapter

axelsrz:
- Added on_turn_error functionality to bot_adapter with updates to all related samples 
- Ported bot assert
- Added some missing annotations.

*This PR address issue #167


libraries/botbuilder-core/botbuilder/core/bot_adapter.py
```diff
@@ -48,12 +50,25 @@ def use(self, middleware):
         """
         self._middleware.use(middleware)
 
-    async def run_middleware(self, context: TurnContext, callback: Callable=None):
+    async def run_middleware(self, context: TurnContext, callback: Callable[[TurnContext], Awaitable]= None):
```
cleemullins: I don't understand the drift here from the C# codebase. The C# has [RunPipelineAsync](https://github.com/microsoft/botbuilder-dotnet/blob/ccf1643e3168a395f327d9eb3b6c351fa2855493/libraries/Microsoft.Bot.Builder/BotAdapter.cs#L158). 

I'm not clear why the name here is different, as it's going to cause a documentation confusion.

libraries/botbuilder-core/botbuilder/core/bot_assert.py
```diff
@@ -0,0 +1,63 @@
+from typing import List
```
cleemullins: Copyright / License statement is missing.

libraries/botbuilder-core/botbuilder/core/bot_adapter.py
axelsrz: This same function in JS is called runMiddleware and does functionally the same. Which one should I follow?

cleemullins: I would stick with whichever codebase you've bene (mostly) modeling the Python SDK after. I have no strong preference.

axelsrz: In previous versions of the Python SDK this function was named after JS. Ultimately I decided to change it given that I'm basing the code for the _most part_ in C# (class names, and structure-wise as well)

libraries/botbuilder-core/botbuilder/core/bot_assert.py
axelsrz: Done, actually I just realized that this class was duplicate but this version follows the correct way of none-checking. Removed dup.

samples/python-flask/13.core-bot/main.py
```diff
@@ -29,6 +29,19 @@
 SETTINGS = BotFrameworkAdapterSettings(app.config['APP_ID'], app.config['APP_PASSWORD'])
 ADAPTER = BotFrameworkAdapter(SETTINGS)
 
+# Catch-all for errors.
+async def on_error(context: TurnContext, error: Exception):
+    # This check writes out errors to console log
+    # NOTE: In production environment, you should consider logging this to Azure
+    #       application insights.
+    print(f'\n [on_turn_error]: { error }')
```
daveta: >print(f'\n [on_turn_error]: { error }') [](start = 4, length = 39)

FYI: I've seen where prints get swallowed (can't recall on linux/windows on flask/django).  I've been getting in the practice of using:
print(f"blah-blah", file=sys.stderr)
Which appears to print slightly more consistently.

daveta:
> gi

whoops

---
Refers to: samples/python-flask/13.core-bot/main.py:11 in a078008. [](commit_id = a0780089d5e7dabe752a753a393bb9ea45e0fda0, deletion_comment = False)

samples/python-flask/13.core-bot/main.py
axelsrz: solved

Issue #140, opened 2019-04-22
MS Teams tenant ID location needs mapping in BotFrameworkAdapter.createConversation


axelsrz:
Parity with js issue https://github.com/Microsoft/botbuilder-js/issues/884

Issue #139, opened 2019-04-22
Add middleware to address transitional location of tenantId field (Parity)


axelsrz:
Parity with [https://github.com/Microsoft/botbuilder-dotnet/pull/1754]

PR #203, opened 2019-06-07
Added middleware to address transitional location of tenant_id field 

axelsrz:
Added middleware to address transitional location of tenant_id field for teams in BotFrameworkAdapter. More testing on adapter pending.

*This PR address issues #139 and #140

### DIFF
--- a/libraries/botbuilder-core/botbuilder/core/bot_adapter.py
+++ b/libraries/botbuilder-core/botbuilder/core/bot_adapter.py
@@ -4,6 +4,7 @@
 from abc import ABC, abstractmethod
 from typing import List, Callable, Awaitable
 from botbuilder.schema import Activity, ConversationReference
+from botframework.connector import Channels
 
 from . import conversation_reference_extension
 from .bot_assert import BotAssert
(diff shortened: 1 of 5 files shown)


libraries/botbuilder-core/botbuilder/core/bot_adapter.py
```diff
@@ -77,6 +78,19 @@ def use(self, middleware):
         BotAssert.context_not_none(context)
 
         if context.activity is not None:
+
+            # Middleware to assign tenant_id from channelData to Conversation.tenant_id.
```
johnataylor: comment "Middleware"

libraries/botbuilder-core/botbuilder/core/bot_adapter.py
```diff
@@ -77,6 +78,19 @@ def use(self, middleware):
...
+            # MS Teams currently sends the tenant ID in channelData and the correct behavior is to expose 
+            # this value in Activity.Conversation.tenant_id.
+            # This code copies the tenant ID from channelData to Activity.Conversation.tenant_id.
+            # Once MS Teams sends the tenant_id in the Conversation property, this middleware can be removed.
+            if (Channels.ms_teams == context.activity.channel_id and context.activity.conversation is not None
```
johnataylor: this should really only be in the path in BotFrameworkAdapter... in p[rocess_activity in the derived class

Issue #206, opened 2019-06-12
Add LuisPredictionOptions to LuisRecognizer.RecognizeAsync()


gabog:
This issue tracks https://github.com/microsoft/botframework-sdk/issues/5405

PR #209, opened 2019-06-14
206 add luis prediction options to recognizer

jspruance:
Fixes: #206

Add ability for user to pass luis prediction options into 'Recognizer.recognize' function. Whatever fields are present should override what was set previously via constructor to 'Recognizer._options', for the duration of the request.

### DIFF
--- a/libraries/botbuilder-ai/botbuilder/ai/luis/luis_recognizer.py
+++ b/libraries/botbuilder-ai/botbuilder/ai/luis/luis_recognizer.py
@@ -302,7 +302,7 @@ class LuisRecognizer(object):
         luis_result: LuisResult = None
 
         if luis_prediction_options:
-            options = self._merge_options(self._options, luis_prediction_options)
+            options = self._merge_options(luis_prediction_options)
         else: 
             options = self._options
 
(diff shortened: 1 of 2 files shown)


libraries/botbuilder-ai/botbuilder/ai/luis/luis_recognizer.py
```diff
@@ -362,3 +370,11 @@ def fill_luis_event_properties(
         )
 
         await turn_context.send_activity(trace_activity)
+
+    def _merge_options(
+        self,
+        userDefinedOptions
```
axelsrz: This might be nitpicking so feel free to ignore it. 
Although this method is "private" I would consider to annotate it in the form:
`user_defined_options: LuisPredictionOptions `

libraries/botbuilder-ai/botbuilder/ai/luis/luis_recognizer.py
```diff
@@ -362,3 +370,11 @@ def fill_luis_event_properties(
...
+    ) -> LuisPredictionOptions:
+        mergedOptions = LuisPredictionOptions()
```
axelsrz: If possible, consider changing this to snake_case :)

libraries/botbuilder-ai/botbuilder/ai/luis/luis_recognizer.py
jspruance: Not at all...great feedback : ) Ok, this i updated.

libraries/botbuilder-ai/botbuilder/ai/luis/luis_recognizer.py
jspruance: agree...updated to snake case

PR #219, opened 2019-06-19
initial django core-bot sample

tsuwandy:


daveta:
@tsuwandy : It might be a good idea if we start out `pylint` clean (with exception of string length, which I regularly disable).  I have to be honest, the code I checked in earlier was not clean.  I just started this with the 201 sample.  I can show you how to enable in vscode.

For the main bot class stuff, you can steal what I did in 201 (which should be the majority of the code).

PR #226, opened 2019-06-23
Add ARM template to flask core-bot

stevengum:
### Changes
- Add ARM template to flask core-bot for provisioning resources on azure
  - This ARM template creates a _Linux_ App Service as Azure recommends using Linux to deploy your Python Apps
  - The template configures the App Service to build the code on the server by default

- Updates `config.py` to import `os` and read environment variables, similar to the usage of `process.env` in the Node.js core-bot samples


samples/python-flask/13.core-bot/config.py
```diff
@@ -2,13 +2,14 @@
...
     """ Bot Configuration """
     PORT = 3978
-    APP_ID = ""
-    APP_PASSWORD = ""
-    LUIS_APP_ID = ""
-    LUIS_API_KEY = ""
+    APP_ID = os.environ.get("MicrosoftAppId", "")
+    APP_PASSWORD = os.environ.get("MicrosoftAppPassword", "")
+    LUIS_APP_ID = os.environ.get("LuisAppId", "")
+    LUIS_API_KEY = os.environ.get("LuisAPIKey", "")
```
daveta: Curious if you've seen this convention documented before?  This seems to be combining configuration from files, with environment variables (and the environment variables don't appear to match the casing or spelling of the variables you reference in the code).

When we initially were discussing this earlier, these were referenced as convention for Python :
- [http://exploreflask.com/en/latest/configuration.html](http://exploreflask.com/en/latest/configuration.html)
- [http://flask.pocoo.org/docs/1.0/config/#configuring-from-environment-variables](http://flask.pocoo.org/docs/1.0/config/#configuring-from-environment-variables)

They do provide examples of environment variables, but don't appear to do indirection through the config object, just directly read into variables.

@axelsrz

samples/python-flask/13.core-bot/config.py
```diff
@@ -2,13 +2,14 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
 
-""" Bot Configuration """
```
daveta: May want to add this back in so pylint doesn't complain.

stevengum: Done!

samples/python-flask/13.core-bot/config.py
stevengum: So the reason I went went the variable keys above is because that's what we use when we provision bots on Azure. We can change the keys right now because we will be using Linux Web Apps instead of Windows Web Apps, but these are the keys we've used across C# and JS to use only one ARM template.

After some preliminary explorations and discussions, I think that the pattern I used above is the simplest way to get a local Python Flask sample bot deployed and running on Azure.

It populates the config with default values, but when deployed to a Linux App Service with the properly configured Application Settings, it reads the environment variables and so the code is ready to use in production.

PR #228, opened 2019-06-26
Tests for activity prompt. 

axelsrz:
Fixes on prompt, prompt_validator_context, activity_prompt and test_adapter

### DIFF
--- a/libraries/botbuilder-dialogs/tests/test_activity_prompt.py
+++ b/libraries/botbuilder-dialogs/tests/test_activity_prompt.py
@@ -11,7 +11,7 @@ from botbuilder.schema import Activity, InputHints, ActivityTypes
 
 from botbuilder.core import ConversationState, MemoryStorage, TurnContext, MessageFactory
 from botbuilder.core.adapters import TestAdapter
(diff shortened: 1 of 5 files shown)


libraries/botbuilder-dialogs/tests/test_activity_prompt.py
```diff
@@ -2,35 +2,203 @@
...
+
 
 class SimpleActivityPrompt(ActivityPrompt):
-    pass
+    def __init__(self, dialog_id: str, validator: Callable[[PromptValidatorContext], bool]):
+        super().__init__(dialog_id, validator)
+
 
 class ActivityPromptTests(aiounittest.AsyncTestCase):
     async def test_does_the_things(self):
```
Zerryth: We should probably actually just remove this test case-- I left it in there, but didn't get around to putting any assertions or anything in there

Zerryth: referencing test called `async def test_does_the_things()`

axelsrz: Solved (:

PR #229, opened 2019-06-26
Fix django async/loop

daveta:
- `async` fix in `aux_func` for Django
- Move `loop` to be global

### DIFF
--- a/samples/django/13.core-bot/bots/views.py
+++ b/samples/django/13.core-bot/bots/views.py
@@ -30,17 +30,17 @@ def messages(request):
 
     activity = Activity().deserialize(body)
     auth_header = request.headers['Authorization'] if 'Authorization' in request.headers else ''
-    loop = asyncio.get_event_loop()
 
     bot_app = apps.get_app_config('bots')
     bot = bot_app.bot
     adapter = bot_app.ADAPTER
 
     async def aux_func(turn_context):
-        asyncio.ensure_future(bot.on_turn(turn_context), loop=loop)
+        await bot.on_turn(turn_context)
+
     try:
-        task = asyncio.ensure_future(adapter.process_activity(activity, auth_header, aux_func), loop=loop)
-        loop.run_until_complete(task)
+        task = asyncio.ensure_future(adapter.process_activity(activity, auth_header, aux_func), loop=bot_app.LOOP)
+        bot_app.LOOP.run_until_complete(task)
         return HttpResponse(status=201)
     except Exception as exception:
         raise exception
--- a/samples/django/13.core-bot/bots/views.py
+++ b/samples/django/13.core-bot/bots/views.py
@@ -33,14 +33,15 @@ def messages(request):
 
     bot_app = apps.get_app_config('bots')
     bot = bot_app.bot
+    loop = bot_app.LOOP
     adapter = bot_app.ADAPTER
 
     async def aux_func(turn_context):
         await bot.on_turn(turn_context)
 
     try:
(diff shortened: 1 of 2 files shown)


samples/django/13.core-bot/bots/views.py
```diff
@@ -30,17 +30,17 @@ def messages(request):
 
     activity = Activity().deserialize(body)
     auth_header = request.headers['Authorization'] if 'Authorization' in request.headers else ''
-    loop = asyncio.get_event_loop()
```
tsuwandy: loop = bot_app.LOOP

samples/django/13.core-bot/bots/views.py
```diff
@@ -30,17 +30,17 @@ def messages(request):
...
     adapter = bot_app.ADAPTER
 
     async def aux_func(turn_context):
-        asyncio.ensure_future(bot.on_turn(turn_context), loop=loop)
+        await bot.on_turn(turn_context)
+
     try:
-        task = asyncio.ensure_future(adapter.process_activity(activity, auth_header, aux_func), loop=loop)
-        loop.run_until_complete(task)
+        task = asyncio.ensure_future(adapter.process_activity(activity, auth_header, aux_func), loop=bot_app.LOOP)
```
tsuwandy: task = asyncio.ensure_future(adapter.process_activity(activity, auth_header, aux_func), loop=loop)

samples/django/13.core-bot/bots/views.py
```diff
@@ -30,17 +30,17 @@ def messages(request):
...
+        bot_app.LOOP.run_until_complete(task)
```
tsuwandy: loop.run_until_complete(task)

samples/django/13.core-bot/bots/views.py
daveta: Great idea!

samples/django/13.core-bot/bots/views.py
daveta: Thanks, fixed

samples/django/13.core-bot/bots/views.py
daveta: Thanks, fixed

viperfx:
hi there - I am not able to find django samples in this repo or the sample python repo (https://github.com/microsoft/BotBuilder-Samples/tree/master/samples)

Have they been moved elsewhere?

Any guidance on how to use async method, django and bot framework would be great.

axelsrz:
Hello @viperfx , given that the framework heavily relies on asyncio syntax we no longer support django. We recommend you look into sync frmeworks like aiohttp or tornado.

viperfx:
Hey @axelsrz thanks for the reply. Given that django is one of the most popular frameworks in python and it's very unlikely developers are going to run two different frameworks and libraries side by side for applications, it really would benefit the adoption of SDK framework to support what is currently the state of python ecosystem. The python3 adoption is good, but async I would say is still very early. Lot of database and ORM adapters do not support async right now, and still being actively worked on. All these pieces need to come together to make an application to serve end users.

Django has recently added support for Async - https://docs.djangoproject.com/en/3.0/topics/async/

But as you can see, there is a phased approach for adding full async support.

I would really encourage another look and adding some samples around how a non-async codebase in django could use bot framework, e.g. pre django 3.0, and if that is not possible, then adding a sample for django 3.0 onnwards with bot framework.

We have an existing solution in Teams that uses our internal SDK with REST APIs, but we are interested in using the bot framework too with our existing code base.

PR #231, opened 2019-06-27
tests for tokenizer

axelsrz:


Zerryth:
Definitely get Chris's blessing, since he knows more about the unicode rabbit hole :)
But great job on getting the unit tests in and correcting the behavior to be language-specific to python

Could you change one small typo that I included when I was transliterating from C#? (oopsy):
LN 31 inside the while loop should say "both the" instead :)

libraries/botbuilder-dialogs/tests/choices/test_choice_tokenizer.py
```diff
@@ -0,0 +1,57 @@
+import aiounittest
```
carlosscastro: We should add copyright headers on all files

PR #260, opened 2019-07-16
ChoicePrompt properly uses ListStyle that's specified

### LINKED ISSUE #239: ChoiceFactory forces usage of hardcoded ChoiceFactoryOptions; can therefore only use en-us locale
## Version
python v4.4

## Describe the bug
* `ChoiceFactory.inline()` is broken, in that it does not use the passed in `options` parameter, and only uses `opt`, the hard-coded "or" separators
* this in turn means that English is the only locale that prompts are able to use
* also, due to https://github.com/microsoft/botbuilder-python/issues/240, we can't move away from inline styling, therefore we can't move away from `en-us` locale

## Expected behavior
To use the locale specified, and not only stick to `en-us`

### LINKED ISSUE #240: Prompt styling always sticks to ListStyle.in_line, even when specifying a different style
## Version
python v4.4 (preview 1)

## Describe the bug
* create `ChoicePrompt`
* set `ChoicePrompt.style` to anything besides `ListStyle.in_line` or `ListStyle.auto` (which is `in_line` anyways)
    * i.e. `ListStyle.list_style`, `ListStyle.none`, etc.
* bot response sent sticks to `in_line` rather than respecting specified style

## Expected behavior
* to use the styling specified, not just `in_line`

[bug]

Zerryth:
Fixes #239 , #240 

### Changes
* `ChoicePrompt._default_choice_options` now specifies **`inline_or_more`** for the `ChoiceFactoryOptions` in the appropriate culture

* `Prompt` was previously trying to use `Dict.get()` method, but with the enum class as the key instead of the _**value**_ that the enum equated to, making all styles go to "default" `ListStyle` previously, when key was not found (in this case `in_line` styling)
  * now properly uses enum value as key
* rearranged `switcher` variable to properly map in the same order as the values in `ListStyle` enum class, to apply correct styling
* `ListStyle.none` also now returns as an `ActivityType.message`, whereas before it was `None`
* updated unit tests to not only check for `in_line` styling

### DIFF
--- a/libraries/botbuilder-dialogs/tests/test_choice_prompt.py
+++ b/libraries/botbuilder-dialogs/tests/test_choice_prompt.py
@@ -25,18 +25,18 @@ from botbuilder.dialogs.prompts import (
 from botbuilder.schema import Activity, ActivityTypes
 
 _color_choices: List[Choice] = [
-    Choice(value="red"),
-    Choice(value="green"),
-    Choice(value="blue"),
+    Choice(value='red'),
+    Choice(value='green'),
+    Choice(value='blue'),
 ]
 
-_answer_message: Activity = Activity(text="red", type=ActivityTypes.message)
-_invalid_message: Activity = Activity(text="purple", type=ActivityTypes.message)
+_answer_message: Activity = Activity(text='red', type=ActivityTypes.message)
+_invalid_message: Activity = Activity(text='purple', type=ActivityTypes.message)
 
 
 class ChoicePromptTest(aiounittest.AsyncTestCase):
     def test_choice_prompt_with_empty_id_should_fail(self):
-        empty_id = ""
+        empty_id = ''
 
         with self.assertRaises(TypeError):
             ChoicePrompt(empty_id)
@@ -56,11 +56,11 @@ class ChoicePromptTest(aiounittest.AsyncTestCase):
             if results.status == DialogTurnStatus.Empty:
                 options = PromptOptions(
                     prompt=Activity(
-                        type=ActivityTypes.message, text="Please choose a color."
+                        type=ActivityTypes.message, text='Please choose a color.'
                     ),
                     choices=_color_choices,
                 )
-                await dc.prompt("ChoicePrompt", options)
+                await dc.prompt('ChoicePrompt', options)
             elif results.status == DialogTurnStatus.Complete:
                 selected_choice = results.result
                 await turn_context.send_activity(selected_choice.value)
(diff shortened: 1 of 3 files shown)


libraries/botbuilder-dialogs/tests/test_choice_prompt.py
```diff
@@ -267,11 +267,8 @@ def test_choice_prompt_with_none_id_should_fail(self):
         dialogs.add(choice_prompt)
 
         step1 = await adapter.send(Activity(type=ActivityTypes.message, text="Hello"))
-        # TODO ChoiceFactory.inline() is broken, where it only uses hard-coded English locale.
-        # commented out the CORRECT assertion below, until .inline() is fixed to use proper locale
-        # step2 = await step1.assert_reply('Please choose a color. (1) red, (2) green, o (3) blue')
         step2 = await step1.assert_reply(
-            "Please choose a color. (1) red, (2) green, or (3) blue"
+            "Please choose a color. (1) red, (2) green, o (3) blue"
```
axelsrz: Is this 'o' expected behavior?

libraries/botbuilder-dialogs/tests/test_choice_prompt.py
```diff
@@ -318,11 +315,8 @@ def test_choice_prompt_with_none_id_should_fail(self):
         step1 = await adapter.send(
             Activity(type=ActivityTypes.message, text="Hello", locale=Culture.Spanish)
         )
-        # TODO ChoiceFactory.inline() is broken, where it only uses hard-coded English locale.
-        # commented out the CORRECT assertion below, until .inline() is fixed to use proper locale
-        # step2 = await step1.assert_reply('Please choose a color. (1) red, (2) green, o (3) blue')
         step2 = await step1.assert_reply(
-            "Please choose a color. (1) red, (2) green, or (3) blue"
+            'Please choose a color. (1) red, (2) green, o (3) blue'
```
axelsrz: Because we're adopting black code formatter soon, could you change the single quote marks for " (double) for all the strings please?

libraries/botbuilder-dialogs/tests/test_choice_prompt.py
Zerryth: Yes, 'o' is expected in this particular test case, because it's testing 'es-es' locale :)

libraries/botbuilder-dialogs/tests/test_choice_prompt.py
Zerryth: Oh good! I like single quotes better for strings anyways :)

libraries/botbuilder-dialogs/tests/test_choice_prompt.py
Zerryth: Although it admittedly does look odd that most of the sentence is in English, except the "inline or more separator"

libraries/botbuilder-dialogs/tests/test_choice_prompt.py
Zerryth: Done! Changed double quotes to single quotes in ChoicePrompt & Prompt classes, and also ChoicePrompt unit tests

PR #279, opened 2019-08-01
NumberPrompt Locale Implementation Fix

### LINKED ISSUE #269: NumberPrompt Locale Not Fully Implemented
## Version
v4.5

## Describe the bug
Found this bug while investigating for parity with regards to this [`NumberPrompt` bug filed in the dotnet repo](https://github.com/microsoft/botbuilder-dotnet/issues/2288)
* in constructor, `default_locale` attribute is set, but never used in `NumberPrompt`'s implementation (not in `on_prompt()` nor `on_recognize()`
* `on_recognize()` does allow you to pass in `locale` via `Activity`, however locale will not be used if only `default_locale` is specified
* "`English`" is used as string to specify locale, when we should be using the constants provided by the python recognizers-text repo

* Separately, there's definitely a lack of unit test coverage for this feature (only 2 tests written)

## Expected behavior
* implement use of `default_locale`
* implement use of recognizers-text constants to specify locale

[bug]

Zerryth:
fixes #269 
___

### Changes

* `NumberPrompt.on_recognize()` now uses default locale if it is set
* `NumberPrompt` uses constant from recognizer-text repo now for English culture instead of hard-coding "`English`" (uses `en-us` now)
* Uses Babel to parse string to number value from the `ModelResult` returned from calling recognize_number() from recognizers-text


Zerryth:
@axelsrz this fixes the locale/culture issue not working in `NumberPrompt` for parity with C#

Separate from the culture issue, found a different bug regarding the validator in `NumberPrompt`, which can be taken care of in a separate PR. Filed #281

libraries/botbuilder-dialogs/botbuilder/dialogs/prompts/number_prompt.py
```diff
@@ -30,9 +41,11 @@ def __init__(self, dialog_id: str, validator: object, default_locale: str):
 
         if is_retry and options.retry_prompt is not None:
             turn_context.send_activity(options.retry_prompt)
-        else:
-            if options.prompt is not None:
-                await turn_context.send_activity(options.prompt)
+        # else:
```
axelsrz: Could you remove the commented code?

Zerryth: Oops. How'd that get back in there. Gone! :)

PR #285, opened 2019-08-09
Message reaction in activity handler

### LINKED ISSUE #243: Add support for Message Reactions to ActivityHandler
ActivityHandler should be extended to include MessageReactions. This has now been added to the C# and The JavaScript.

Here is a pointer to the JavaScript implementation:

https://github.com/microsoft/botbuilder-js/pull/1038

axelsrz:
fixes #243

### DIFF
--- a/libraries/botbuilder-core/botbuilder/core/activity_handler.py
+++ b/libraries/botbuilder-core/botbuilder/core/activity_handler.py
@@ -1,7 +1,9 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
 
-from botbuilder.schema import ActivityTypes, ChannelAccount
+from typing import List
+
+from botbuilder.schema import ActivityTypes, ChannelAccount, MessageReaction
 from .turn_context import TurnContext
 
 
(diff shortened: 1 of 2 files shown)


libraries/botbuilder-core/botbuilder/core/activity_handler.py
```diff
@@ -27,14 +29,15 @@ class ActivityHandler:
...
+            await self.on_message_reaction_activity(turn_context)
         elif turn_context.activity.type == ActivityTypes.event:
             await self.on_event_activity(turn_context)
         else:
             await self.on_unrecognized_activity_type(turn_context)
 
-    async def on_message_activity(  # pylint: disable=unused-argument
-        self, turn_context: TurnContext
-    ):
+    @abstractmethod
```
johnataylor: Why is this abstract? wouldn't that mean you _have to_ implement it? I don't think that is the semantic on the other implementations. (If it is, that that seems like a bug)

axelsrz: Fixed

axelsrz:
@johnataylor ABC removed

PR #290, opened 2019-08-09
PrivateConversationState 

### LINKED ISSUE #145: PrivateConversationState
Feature parity with https://github.com/Microsoft/botbuilder-dotnet

axelsrz:
PrivateConversationState implemented with basic unit test, more testing pending.

fixes #145

### DIFF
--- a/libraries/botbuilder-core/botbuilder/core/private_conversation_state.py
+++ b/libraries/botbuilder-core/botbuilder/core/private_conversation_state.py
@@ -1,0 +1,36 @@
+from .bot_state import BotState
+from .turn_context import TurnContext
+from .storage import Storage
+
+
+class PrivateConversationState(BotState):
+    def __init__(self, storage: Storage, namespace: str = ""):
+        async def aux_func(context: TurnContext) -> str:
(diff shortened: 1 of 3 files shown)


libraries/botbuilder-core/botbuilder/core/private_conversation_state.py
```diff
@@ -0,0 +1,36 @@
...
+        if (
+            activity
+            and activity.from_property
+            and activity.from_property.id is not None
+        ):
+            user_id = activity.from_property.id
+        else:
+            raise Exception("missing activity.from_property.id")
+
+        return f"{channel_id}/conversations/{ conversation_id }/users/{ user_id }/{ self.namespace }"
```
johnataylor: The c# doesn't have the /{ self.namespace } part???

axelsrz: But JS does

johnataylor: so we are saying this is a bug in the C# can you open an issue?

PR #293, opened 2019-08-12
Telemetry Middleware

daveta:
Initial port of Telemetry Middleware.


axelsrz:
Could you format the code to be black format compliant? 
`pip install black`

On the root of the project:
`black libraries`

daveta:
Nice tool!

PR #294, opened 2019-08-12
Axsuarez/blob storage

### LINKED ISSUE #235: Porting of BlobStorage in botbuilder azure
port functionality for storing in Blob storage

axelsrz:
fixes #235

### DIFF
--- a/libraries/botbuilder-azure/botbuilder/azure/blob_storage.py
+++ b/libraries/botbuilder-azure/botbuilder/azure/blob_storage.py
@@ -1,7 +1,7 @@
 import json
-from typing import List
+from typing import Dict, List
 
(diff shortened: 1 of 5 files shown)


libraries/botbuilder-azure/botbuilder/azure/blob_storage.py
```diff
@@ -0,0 +1,102 @@
...
+            )
+
+        self.client = client
+        self.settings = settings
+
+    async def read(self, keys: List[str]) -> Dict[str, object]:
+        if not keys:
+            raise Exception("Please provide at least one key to read from storage.")
+
+        self.client.create_container(self.settings.container_name)
```
johnataylor: no async lib???

can you open an issue to track the fact that we would like an async version

PR #316, opened 2019-09-17
[QnAMaker] GetAnswerRaw added

trojenguri:
Added Active learning enabled flag support
Tests added for that.
[dotnet PR]: https://github.com/microsoft/botbuilder-dotnet/pull/2512
[node PR]: https://github.com/microsoft/botbuilder-js/pull/1172


libraries/botbuilder-ai/botbuilder/ai/qna/models/query_results.py
```diff
@@ -9,11 +9,22 @@
 class QueryResults:
     """ Contains answers for a user query. """
 
-    def __init__(self, answers: List[QueryResult]):
+    _attribute_map = {
```
axelsrz: You only need to add the attribute map if you're gonna inherit from msrest.serialization.Model which I suspect you don't need.

In the rest of the sdk we sometimes inherit from msrest because of the convenience that methods to convert from json provide. If you don't do that with this model consider removing it.

libraries/botbuilder-ai/botbuilder/ai/qna/models/query_results.py
```diff
@@ -9,11 +9,22 @@
...
+    ):
         """
         Parameters:
         -----------
 
         answers: The answers for a user query.
+
+        active_learning_enabled: The active learning enable flag.
         """
+        super(QueryResults, self).__init__(**kwargs)
```
axelsrz: The same is true for the **kwargs argument and the super() initialization call, if you don't inherit from msrest model they are not necessary.

libraries/botbuilder-ai/botbuilder/ai/qna/models/query_results.py
trojenguri: We would need attrubute map to format key and attribute values. And we are reading these values from rest apis

axelsrz: I see, I guess the fix here was to inherit from msrest Model

PR #326, opened 2019-09-30
Adding Recognizer interface to LuisRecognizer

### LINKED ISSUE #254: LuisRecognizer is missing use of interface
## Version
What package version of the SDK are you using.

## Describe the bug
Give a clear and concise description of what the bug is.

## Expected behavior
Give a clear and concise description of what you expected to happen.

## Additional context

[bug]

axelsrz:
Fixes #254

### DIFF
--- a/libraries/botbuilder-ai/botbuilder/ai/luis/luis_recognizer.py
+++ b/libraries/botbuilder-ai/botbuilder/ai/luis/luis_recognizer.py
@@ -8,7 +8,13 @@ from azure.cognitiveservices.language.luis.runtime import LUISRuntimeClient
 from azure.cognitiveservices.language.luis.runtime.models import LuisResult
 from msrest.authentication import CognitiveServicesCredentials
 
-from botbuilder.core import BotAssert, IntentScore, RecognizerResult, TurnContext
+from botbuilder.core import (
+    BotAssert,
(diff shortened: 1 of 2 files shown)


libraries/botbuilder-ai/botbuilder/ai/luis/luis_recognizer.py
```diff
@@ -8,15 +8,21 @@
...
+)
 from botbuilder.schema import ActivityTypes
 
 from . import LuisApplication, LuisPredictionOptions, LuisTelemetryConstants
 from .activity_util import ActivityUtil
 from .luis_util import LuisUtil
 
 
-class LuisRecognizer:
+class LuisRecognizer(Recognizer):
```
daveta: Hmm - not sure if this is for Luis V3 or old one, but looking at C#, there's another level of TelemetryRecognizer if you want to be faithful to that..

axelsrz: Old one. Yes, but that interface only provide more method overloading (not supported in python) and attributes (not supported in the ABC pattern), making it redundant.

The idea here is to provide an interface that can be used at testing, where this ABC is sufficient.

PR #375, opened 2019-10-29
Core bot generator

axelsrz:


generators/app/templates/core/{{cookiecutter.bot_name}}/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Prerequisites
+
+This sample **requires** prerequisites in order to run.
+
+### Overview
+
+This bot uses [LUIS](https://www.luis.ai), an AI based cognitive service, to implement language understanding.
+
+### Install Python 3.7
```
daveta: >Python 3.7 [](start = 12, length = 10)

I'm still not sure if we're 3.7 or 3.6 here. ;)

axelsrz: Solved

PR #381, opened 2019-10-30
Added 47.inspection

tracyboehrer:
Also corrected README in 45.state-management


samples/47.inspection/app.py
```diff
@@ -0,0 +1,113 @@
...
+            timestamp=datetime.utcnow(),
+            type=ActivityTypes.trace,
+            value=f"{error}",
+            value_type="https://www.botframework.com/schemas/error"
+        )
+        # Send a trace activity, which will be displayed in Bot Framework Emulator
+        await context.send_activity(trace_activity)
+        
+    # Clear out state
+    await CONVERSATION_STATE.delete(context)
```
axelsrz: Here we could take conversation_state directly from self, we could discuss which one is best but if we go by this approach (taking the state from a higher context) we should get rid of self and modify line 62 to not convert the function to a MethodType. 

Please tell me what you think

tracyboehrer: Check out my change.  If acceptable, I'l circle back and change the other samples.

PR #421, opened 2019-11-08
Added 15.handling attachments

tracyboehrer:


samples/15.handling-attachments/app.py
```diff
@@ -0,0 +1,83 @@
...
+            name="on_turn_error Trace",
+            timestamp=datetime.utcnow(),
+            type=ActivityTypes.trace,
+            value=f"{error}",
+            value_type="https://www.botframework.com/schemas/error"
+        )
+        # Send a trace activity, which will be displayed in Bot Framework Emulator
+        await context.send_activity(trace_activity)
+
+ADAPTER.on_turn_error = MethodType(on_error, ADAPTER)
```
axelsrz: We should update this to be a simple function not an instance method

samples/15.handling-attachments/bots/attachments_bot.py
```diff
@@ -0,0 +1,216 @@
...
+            conversation_id,
+            AttachmentData(
+                name="architecture-resize.png",
+                original_base64=image_data,
+                type="image/png"
+            )
+        )
+
+        base_uri: str = connector.config.base_url
+        attachment_uri = \
```
axelsrz: Could you use parenthesis instead of \ ? 
Just a simple convention of the SDK.

PR #423, opened 2019-11-11
Conversation Update Scenario

Virtual-Josh:
Adding the conversation update scenario


libraries/botbuilder-schema/botbuilder/schema/teams/channel_info.py
```diff
@@ -0,0 +1,29 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+#
+# Code generated by Microsoft (R) AutoRest Code Generator.
+# Changes may cause incorrect behavior and will be lost if the code is
+# regenerated.
+
+
+class ChannelInfo:
```
axelsrz: This is a matter of style but we're trying to restrict the use of properties to only use them when some . validation is needed. If the attributes are going to be accessed normally please use plane attr.

libraries/botbuilder-schema/botbuilder/schema/teams/notification_info.py
```diff
@@ -0,0 +1,20 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+#
+# Code generated by Microsoft (R) AutoRest Code Generator.
+# Changes may cause incorrect behavior and will be lost if the code is
+# regenerated.
+
+
+class NotificationInfo:
```
axelsrz: Same comment here about attrs vs properties

Virtual-Josh:
Finishing in other PR since I don't want to cherry pick due to the A

PR #437, opened 2019-11-15
Axsuarez/skills first iteration

### LINKED ISSUE #348: Implement remote skill invocation components (Python)
Implements https://github.com/microsoft/botframework-sdk/issues/5584

axelsrz:
fixes #348


libraries/botbuilder-core/botbuilder/core/integration/bot_framework_http_client.py
```diff
@@ -0,0 +1,115 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
stevengum: I believe this should just be in `core/`, there isn't a parallel story for this class here in Python compared to C#.

PR #446, opened 2019-11-22
_dialog_stack to dialog_stack

mdrichardson:
Prevents the following error:

```python
>       self._stack = state.dialog_stack
E       AttributeError: 'dict' object has no attribute 'dialog_stack'
```

dialog_context was errantly referencing `state.dialog_stack`, which didn't exist because dialog_state has it as `self._dialog_stack`. Since it should be public, this PR changes references of `_dialog_stack` to `dialog_stack` and removes the now-unnecessary `@property`

Note: As far as I can tell `dialog_stack` is only referenced in `dialog_context` and `dialog_state`, so this should cover all necessary changes.


mdrichardson:
Not a bug here. The "bug" I was experiencing was due to a Cosmos module I was writing that wasn't serializing/deserializing appropriately

PR #447, opened 2019-11-26
added claims validator

axelsrz:


axelsrz:
> preemptive approval, spoke with @axelsrz about fixing the claims bit.

Fixed

PR #450, opened 2019-11-26
Fix compute_hash

mdrichardson:
With `compute_hash` as:

```python
def compute_hash(self, obj: object) -> str:
    # TODO: Should this be compatible with C# JsonConvert.SerializeObject ?
    return str(obj)
```

...the hashes always return as something like "DialogStack at memoryAddressXYZ". If the DialogStack changes (dialogs added/removed), the memory address doesn't change, so the hash appears to be the same and changes are not written.

With:

```python
def compute_hash(self, obj: object) -> str:
    return str(Pickler().flatten(obj))
```

...the whole DialogStack would get serialized and changes are correctly detected.

---

This wasn't detected because `MemoryStorage` manipulates the objects in memory, anyway and doesn't actually *need* to call `MemoryStorage.write()`. With persistent storage, write wasn't being called enough and things like ConversationState weren't actually updating, preventing users from progressing through WaterfallDialogs.


libraries/botbuilder-core/botbuilder/core/bot_state.py
```diff
@@ -24,8 +25,7 @@ def is_changed(self) -> bool:
         return self.hash != self.compute_hash(self.state)
 
     def compute_hash(self, obj: object) -> str:
-        # TODO: Should this be compatible with C# JsonConvert.SerializeObject ?
-        return str(obj)
+        return str(Pickler().flatten(obj))
```
axelsrz: Great catch! This is so important that I'm surprised that didn't blow up earlier

PR #453, opened 2019-11-26
Axsuarez/skills lower layers

axelsrz:


axelsrz:
Closing given that we're merging #464 which has these changes

PR #454, opened 2019-11-26
Merging Teams in so far

Virtual-Josh:


libraries/botbuilder-core/botbuilder/core/teams/teams_activity_handler.py
```diff
@@ -0,0 +1,406 @@
...
+)
+from botframework.connector import Channels
+
+
+class TeamsActivityHandler(ActivityHandler):
+    async def on_turn(self, turn_context: TurnContext):
+        if turn_context is None:
+            raise TypeError("ActivityHandler.on_turn(): turn_context cannot be None.")
+
+        if hasattr(turn_context, "activity") and turn_context.activity is None:
```
axelsrz: change to if not getattr(turn_context, "activity", None)

libraries/botbuilder-core/botbuilder/core/teams/teams_activity_handler.py
```diff
@@ -0,0 +1,406 @@
...
+            raise TypeError(
+                "ActivityHandler.on_turn(): turn_context must have a non-None activity."
+            )
+
+        if (
+            hasattr(turn_context.activity, "type")
```
axelsrz: same thing here

libraries/botbuilder-core/botbuilder/core/teams/teams_activity_handler.py
```diff
@@ -0,0 +1,406 @@
...
+
+    async def on_teams_members_removed_dispatch_activity(  # pylint: disable=unused-argument
+        self,
+        members_removed: [ChannelAccount],
+        team_info: TeamInfo,
+        turn_context: TurnContext,
+    ):
+        teams_members_removed = []
+        for member in members_removed:
+            new_account_json = member.__dict__
```
axelsrz: do not serialize like this, use the .serialize() method that all msrest.serialization.Model have

libraries/botbuilder-core/botbuilder/core/teams/teams_activity_handler.py
```diff
@@ -0,0 +1,406 @@
...
+        self, team_info: TeamInfo, turn_context: TurnContext
+    ):
+        return  # Task.CompleteTask
+
+    @staticmethod
+    def _create_invoke_response(body: object = None) -> InvokeResponse:
+        return InvokeResponse(status=int(HTTPStatus.OK), body=body)
+
+
+class _InvokeResponseException(BaseException):
```
axelsrz: Inherit from Exception

libraries/botbuilder-core/botbuilder/core/teams/teams_activity_handler.py
```diff
@@ -0,0 +1,406 @@
...
+                        name = member.name,
+                        aad_object_id = member.aad_object_id,
+                        role = member.role
+                        )
+                    team_members_added.append(newTeamsChannelAccount)
+
+        return await self.on_teams_members_added_activity(teams_members_added, team_info, turn_context)
+        """
+        for member in members_added:
+            new_account_json = member.__dict__
```
axelsrz: same serialization comment here

libraries/botbuilder-core/botbuilder/core/teams/teams_activity_handler.py
```diff
@@ -0,0 +1,406 @@
...
+            if invoke_response and not turn_context.turn_state.get(
+                BotFrameworkAdapter._INVOKE_RESPONSE_KEY  # pylint: disable=protected-access
+            ):
+                await turn_context.send_activity(
+                    Activity(value=invoke_response, type=ActivityTypes.invoke_response)
+                )
+            return
+
+        await super().on_turn(turn_context)
+        return
```
axelsrz: don't need return here

libraries/botbuilder-schema/botbuilder/schema/teams/teams_channel_account.py
```diff
@@ -0,0 +1,31 @@
...
+    def __init__(
+        self,
+        id="",
+        name="",
+        aad_object_id="",
+        role="",
+        given_name="",
+        surname="",
+        email="",
+        userPrincipalName="",
```
tracyboehrer: Is "userPrincipalName" needed?  Why not "user_principal_name"?

libraries/botbuilder-schema/botbuilder/schema/teams/teams_channel_data.py
```diff
@@ -0,0 +1,30 @@
...
+# regenerated.
+
+from botbuilder.schema.teams import ChannelInfo, TeamInfo, NotificationInfo, TenantInfo
+
+
+class TeamsChannelData:
+    def __init__(
+        self,
+        channel: ChannelInfo = None,
+        eventType="",
```
tracyboehrer: Why not "event_type" instead of "eventType"?

Virtual-Josh: Ahh the comment on this one must have been lost. I'll re-add it.

It's this way because I'm using the flat map method (**) to create objects based on the JSON that's coming in on the activity. The name of the field in the JSON is eventType so that's what it has to be here. IIRC the property on the class is correctly cased.

libraries/botbuilder-schema/botbuilder/schema/teams/teams_channel_account.py
Virtual-Josh: See comment below

libraries/botbuilder-core/botbuilder/core/teams/teams_activity_handler.py
```diff
@@ -0,0 +1,404 @@
...
+            if turn_context.activity.name == "composeExtension/querySettingUrl":
+                return self._create_invoke_response(
+                    await self.on_teams_messaging_extension_configuration_query_settings_url_activity(
+                        turn_context, turn_context.activity.value
+                    )
+                )
+
+            if turn_context.activity.name == "composeExtension/setting":
+                await self.on_teams_messaging_extension_configuration_setting_activity(
+                    turn_context, turn_context.turn_context.activity.value
```
tracyboehrer: turn_context.turn_context.activity.  I don't think this is correct?

libraries/botbuilder-core/botbuilder/core/teams/teams_activity_handler.py
```diff
@@ -0,0 +1,404 @@
...
+            new_account_json = member.seralize()
+            del new_account_json["additional_properties"]
+            member = TeamsChannelAccount(**new_account_json)
+        return await self.on_teams_members_added_activity(members_added, turn_context)
+
+    async def on_teams_members_added_activity(
+        self, teams_members_added: [TeamsChannelAccount], turn_context: TurnContext
+    ):
+        for member in teams_members_added:
+            member = ChannelAccount(member)
```
tracyboehrer: What is this doing?

libraries/botbuilder-core/botbuilder/core/teams/teams_activity_handler.py
```diff
@@ -0,0 +1,404 @@
...
+            teams_members_removed.append(TeamsChannelAccount(**new_account_json))
+
+        return await self.on_teams_members_removed_activity(
+            teams_members_removed, turn_context
+        )
+
+    async def on_teams_members_removed_activity(
+        self, teams_members_removed: [TeamsChannelAccount], turn_context: TurnContext
+    ):
+        members_removed = [ChannelAccount(i) for i in teams_members_removed]
```
tracyboehrer: Is the cast needed?  TeamsChannelAccount is a ChannelAccount already.

libraries/botbuilder-core/botbuilder/core/teams/teams_activity_handler.py
```diff
@@ -0,0 +1,404 @@
...
+        self, channel_info: ChannelInfo, team_info: TeamInfo, turn_context: TurnContext
+    ):
+        return  # Task.CompleteTask
+
+    async def on_teams_channel_renamed_activity(  # pylint: disable=unused-argument
+        self, channel_info: ChannelInfo, team_info: TeamInfo, turn_context: TurnContext
+    ):
+        return  # Task.CompleteTask
+
+    async def on_teams_team_reanamed_async(  # pylint: disable=unused-argument
```
tracyboehrer: Probably can push this to the issue where you rename the methods.  But "async" wouldn't be consistent?  Also, spelling error.

libraries/botbuilder-core/botbuilder/core/teams/teams_activity_handler.py
Virtual-Josh: it's doing a downcast from TeamsChannelAccount to a normal ChannelAccount. In short Teams wanted to have us send around TeamsChannelAccount objects which are slightly richer in terms of data. When I wrote the conversation-update bot I didn't have all of the supporting classes so I wrote something to unblock myself. I'll fix this part after the PR gets in since I'll need to update the unit tests around this one as well.

libraries/botbuilder-core/botbuilder/core/teams/teams_activity_handler.py
Virtual-Josh: It should be needed, TeamsChannelAccount is a richer ChannelAccount. Since we are calling the base on_members_removed which takes a list of ChannelAccount to make it happy.

tracyboehrer: Technically it doesn't hurt anything.  But a cast of a TeamsChannelAccount to a ChannelAccount is still a TeamsChannelAccount.  If you were to compare the contents of the list, you'd see they were the same objects:  A TeamsChannelAccount.  C# has to cast because of strong typing.

What's also interesting is ActivityHandler.on_members_added_activity is declared as wanting a ChannelAccount, not a list.  Python won't care, it will run the same.  But in C# it is declared as requiring a List.

As a side note, ActivityHandler uses the "_activity" suffix on those methods.  We need to coordinate with Axel on this since John requested the "_activity" be removed on the TeamsActivityHandler.  May not be a change we want to make at this point.

PR #459, opened 2019-11-27
Cosmos partitioned - Parity with C# and Node | Also storage base tests

### LINKED ISSUE #327: Port new CosmosDB Partitioned Storage from C#
A new Cosmos DB storage implementation which utilizes the v3 API and handles the partition key internally has been implemented in C#.

C# implementation is here https://github.com/microsoft/botbuilder-dotnet/pull/2613

mdrichardson:
This is kind of two PRs in one.

## 1

Fixes #327

Add `cosmosdb_paritioned_storage` for parity with both C# and Node

## 2

Parity with: https://github.com/microsoft/botbuilder-js/pull/1371

The Dotnet repo has a [StorageBaseTests class](https://github.com/microsoft/botbuilder-dotnet/blob/master/tests/Microsoft.Bot.Builder.Tests/StorageBaseTests.cs) that all of the Storage classes borrow from to reduce duplicate code and ensure that the Storage classes all pass a base set of tests.

Python SDK doesn't have this. Instead, there's a ton of duplicate code...until now!

## Specific Changes

This doesn't remove any tests from any of the storage providers. Instead, it replaces them with equivalent tests and in some cases, adds a few for tests that probably should have been conducted, but weren't.

  - Adds `storage_base_tests` class to `botbuilder-testing` for parity w/ Dotnet, which Memory, Blob, Cosmos, and CosmosPartitioned storage providers now implement in addition to their own tests (mostly constructor argument checks)
    - Note: In C#/Node, this falls under `botbuilder-core`, but I couldn't seem to make that work in Python without too many additional `__init__`s. Maybe somebody who isn't Python-dumb can help?
  - Structure/layout of Blob, Cosmos, CosmosPartitioned now matches more closely
  - Some NOOP/linting fixes and general code cleanup

## Changes Worth a Closer Look

- All tests have a [`testRan` check](https://github.com/microsoft/botbuilder-python/pull/459/files#diff-ae8e49a23fce4e7d30d1fb6e450af849R64). Is this necessary? I'd seen it used in some dotnet tests and thought it might be worthwhile since there aren't `assert` checks in most of the individual provider's tests.
- MemoryStorage, BlobStorage, and CosmosDbStorage all had areas where they sort of assumed that objects being written were `StoreItems`. [This shouldn't be the case](https://github.com/microsoft/botbuilder-dotnet/pull/236), so [I changed from `item.e_tag` to `if hasattr(item, "e_tag") <use it> else item.get("e_tag")`](https://github.com/microsoft/botbuilder-python/pull/459/files#diff-9b84d2a62f17059a8b809191befe5d0fR77) in a LOT of places.
- I [threw exceptions for some null/None checks](https://github.com/microsoft/botbuilder-python/pull/459/files#diff-9b84d2a62f17059a8b809191befe5d0fR67) to match other SDKs in several places
- MemoryStorage now `deepcopy()`s on write, which I think is related to [this PR](https://github.com/microsoft/botbuilder-python/pull/450)
- In BlobStorage, I removed [this line](https://github.com/microsoft/botbuilder-python/pull/459/files#diff-9b84d2a62f17059a8b809191befe5d0fL105), `item["id"] = blob.name`, in read(), because it would overwrite an item that had that `id` set and didn't seem to make any tests fail when changing it
- This is just generally a fairly big change and has implications on testing validity of future PRs. Worth a pretty close look just for that reason.

### Testing

CoreBot works w/ MemoryStorage:

image

CoreBot works w/ Blob:

image

CoreBot works w/ Cosmos:

image

CoreBot works w/ CosmosPartitioned:

image


EricDahlvang:
We should consider the use of CosmosDbKeyEscape.sanitize_key and whether or not 255 is an actual requirement by the libraries used for this.

libraries/botbuilder-azure/botbuilder/azure/blob_storage.py
```diff
@@ -63,24 +62,31 @@ def __init__(self, settings: BlobStorageSettings):
...
         )
 
-        for name, item in changes.items():
-            e_tag = (
-                None if not hasattr(item, "e_tag") or item.e_tag == "*" else item.e_tag
-            )
-            if e_tag:
-                item.e_tag = e_tag.replace('"', '\\"')
+        for (name, item) in changes.items():
+            e_tag = item.e_tag if hasattr(item, "e_tag") else item.get("e_tag", None)
```
axelsrz: was there any reason to change previous form? This version has more conditionals to achieve the same thing

libraries/botbuilder-azure/botbuilder/azure/blob_storage.py
```diff
@@ -63,24 +62,31 @@ def __init__(self, settings: BlobStorageSettings):
...
+            e_tag = None if e_tag == "*" else e_tag
+            if e_tag == "":
```
axelsrz: if not e_tag

libraries/botbuilder-azure/botbuilder/azure/cosmosdb_storage.py
```diff
@@ -160,20 +160,30 @@ def __init__(
...
+                    if hasattr(change, "e_tag")
+                    else change.get("e_tag", None)
+                )
                 # create the new document
                 doc = {
                     "id": CosmosDbKeyEscape.sanitize_key(key),
                     "realId": key,
                     "document": self.__create_dict(change),
                 }
+                if e_tag == "":
```
axelsrz: if not e_tag

libraries/botbuilder-azure/botbuilder/azure/blob_storage.py
mdrichardson: Mostly so I could add `item.get("e_tag", None)` to account for non-`StoreItem`s. It was starting to look really sloppy and tough to read if I added that into the conditional. Open to other ideas, though.

libraries/botbuilder-azure/botbuilder/azure/blob_storage.py
mdrichardson: This was intentional based on [this test](https://github.com/microsoft/botbuilder-dotnet/blob/master/tests/Microsoft.Bot.Builder.Tests/StorageBaseTests.cs#L155). If the e_tag is None, it should act as wildcard. If the e_tag is specified as an empty string, it shouldn't work.

libraries/botbuilder-azure/botbuilder/azure/cosmosdb_storage.py
mdrichardson: Same as previous. We want to allow e_tag to be None so it can act as a wildcard, but we don't want the e_tag to be blank.

mdrichardson: Same as previous. We want to allow e_tag to be None so it can act as a wildcard, but we don't want the e_tag to be blank.

PR #464, opened 2019-12-02
Axsuarez/protocol test project

axelsrz:


libraries/botbuilder-core/botbuilder/core/bot_framework_adapter.py
```diff
@@ -228,7 +239,9 @@ def __init__(self, settings: BotFrameworkAdapterSettings):
 
         return await self.run_pipeline(context, logic)
 
-    async def authenticate_request(self, request: Activity, auth_header: str):
+    async def authenticate_request(
+        self, request: Activity, auth_header: str
+    ) -> ClaimsIdentity:
```
stevengum: Isn't this a breaking change? You'll have to either create an internal method that returns the ClaimsIdentity in this method, and call the internal method from `process_activity()`.

I've had to do the exact thing in TS, `BotFrameworkAdapter.authenticateRequest()` is `protected` in TS.

stevengum: Nevermind, we're going GA very soon in Python 🚀

PR #496, opened 2019-12-11
updating create conversation to take conversation params

Virtual-Josh:


johnataylor:
Looks like the create_conversation code in Python is not setting the activity type nor the activity name on the event activity that is used to run the middleware.

Specifically the code:

request = TurnContext.apply_conversation_reference(
                Activity(), reference, is_incoming=True
            )

should set the Activity to have type = "event" and name = "CreateConversation"

libraries/botbuilder-core/botbuilder/core/bot_framework_adapter.py
```diff
@@ -157,6 +157,7 @@ def __init__(self, settings: BotFrameworkAdapterSettings):
         self,
```
johnataylor: This function should set the type and name of the event activity used to warm up the middleware.

Virtual-Josh: Updated.

PR #502, opened 2019-12-13
[Teams] Invoke Response serialization

EricDahlvang:
Thanks for your help with this @Virtual-Josh , @axelsrz and @johnataylor


libraries/botbuilder-core/botbuilder/core/bot_framework_adapter.py
```diff
@@ -263,7 +263,11 @@ def __init__(self, settings: BotFrameworkAdapterSettings):
                     teams_channel_data["tenant"]["id"]
                 )
 
-        return await self.run_pipeline(context, logic)
+        pipeline_result = await self.run_pipeline(context, logic)
```
johnataylor: This doesn't look right. the run_pipeline does not return a value.
Take a look at the C# it should mirror that.

EricDahlvang: addressed (we now return the invoke response: .turn_state.get(...).value

PR #514, opened 2019-12-15
explicit instruction for installing dep

hangdong:


tracyboehrer:
Closing.  The samples in the Python repo are in the process of being moved to the Sampels repo.  I have duplicated this change in that repo.

PR #536, opened 2019-12-17
Teams connector client creation cleanup

EricDahlvang:
Access _credentials from turn_context adapter when creating teams connector client


EricDahlvang:
This has already been cherry-picked to 4.7

PR #632, opened 2020-01-22
emolsh/api-ref-docs-componentdialog

emgrol:


mmiele:
It looks good. 
When you say """ (not sure where to put this information) 
    Containers are typically leaf nodes on the stack but the dev is free to push other dialogs …
I would suggest to use this syntax:
.. note::
    Containers are typically leaf nodes on the stack but the dev is free to push other dialogs …
See https://review.docs.microsoft.com/en-us/help/onboard/admin/reference/python/documenting-api?branch=master#note

PR #653, opened 2020-01-28
Adding helper for starting thread in Teams

Virtual-Josh:


libraries/botbuilder-core/botbuilder/core/teams/teams_info.py
```diff
@@ -14,6 +15,37 @@
 
 
 class TeamsInfo:
+    @staticmethod
+    async def send_message_to_teams_channel(
+        turn_context: TurnContext, activity: Activity, teams_channel_id: str
+    ) -> tuple:
```
axelsrz: could you please import Tuple from typing and specify the expected types of the tuple elements

libraries/botbuilder-core/botbuilder/core/teams/teams_info.py
```diff
@@ -14,6 +15,37 @@
...
+            activity=activity,
+        )
+
+        result = await turn_context.adapter.create_conversation(
+            old_ref, TeamsInfo._create_conversation_callback, conversation_parameters
+        )
+        return (result[0], result[1])
+
+    @staticmethod
+    async def _create_conversation_callback(new_turn_context) -> tuple:
```
axelsrz: same here

PR #683, opened 2020-01-31
continue conversation fix for skills

### LINKED ISSUE #606: Channel returns 401 on ContinueConversationAsync calls
## Version
4.7.0

## Describe the bug
See https://github.com/microsoft/botbuilder-dotnet/issues/3245 for details on the issue.

## To Reproduce
Python uses different caching mechanisms than C#, we need to test and see if we can repro the issue here.

axelsrz:
fixes #606

### DIFF
--- a/libraries/botbuilder-core/botbuilder/core/bot_framework_adapter.py
+++ b/libraries/botbuilder-core/botbuilder/core/bot_framework_adapter.py
@@ -222,6 +222,9 @@ class BotFrameworkAdapter(BotAdapter, UserTokenProvider):
         context = TurnContext(self, get_continuation_activity(reference))
         context.turn_state[BOT_IDENTITY_KEY] = claims_identity
         context.turn_state["BotCallbackHandler"] = callback
+        await self._ensure_channel_connector_client_is_created(
+            reference.service_url, claims_identity
+        )
         return await self.run_pipeline(context, callback)
 
     async def create_conversation(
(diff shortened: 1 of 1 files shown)


libraries/botbuilder-core/botbuilder/core/bot_framework_adapter.py
```diff
@@ -965,3 +968,31 @@ def check_emulating_oauth_cards(self, context: TurnContext):
...
+                claims=default_connector_claims, is_authenticated=True
+            )
+
+            await self.create_connector_client(service_url, connector_claims_identity)
+
+        if SkillValidation.is_skill_claim(claims_identity.claims):
+            # Add the channel service URL to the trusted services list so we can send messages back.
+            # the service URL for skills is trusted because it is applied by the
+            # SkillHandler based on the original request received by the root bot
+            MicrosoftAppCredentials.trust_service_url(service_url)
```
stevengum: This is fine as it mirrors the dotnet impl, but last time we spoke in person you were considering following the [JS implementation (js-#1586)](https://github.com/microsoft/botbuilder-js/pull/1586) which places this logic in the SkillHandler class. @gabog indicated he was going to mirror the JS implementation as well regarding this check for R8.

I was curious to know if there have been any updates or changes re: dotnet moving this check to the SkillHandler class.

axelsrz: Oh ok, I was planning to move it at the same time C# does (didn't know that was happening in R8). I do think JS way is better, is just that because I was aligning to C# I wanted to be consistent with the code parity

stevengum: No problem! I just wanted to know if I needed to change JS 😊

PR #732, opened 2020-02-12
AppCredentials object model and update to MSAL

### LINKED ISSUE #320: Auth: Support cert based authentication for parity

tracyboehrer:
Fixes #320 

This implements the new AppCredentials model.
Ditched ADAL in favor of MSAL (as that team states MSAL is the future).

As a side note, we are using the caching in MSAL instead of our own, which simplifies things.


libraries/botframework-connector/botframework/connector/auth/certificate_app_credentials.py
```diff
@@ -0,0 +1,57 @@
...
+    def get_token(self) -> str:
+        """
+        Implementation of AppCredentials.get_token.
+        :return: The access token for the given certificate.
+        """
+
+        # Firstly, looks up a token from cache
+        # Since we are looking for token for the current app, NOT for an end user,
+        # notice we give account parameter as None.
+        auth_token = self.app.acquire_token_silent(self.scopes, account=None)
```
carlosscastro: Did you inspect the code for acquire_token_for_client or try whether acquire_token_for_client caches? In the C# flavor of MSAL, acquire_token_for_client does the caching itself without external logic.

tracyboehrer: It does appear to cache.  The first time through it falls into the fetch.  Next pass through it does not.  With that said, I did not test for expiration.

PR #821, opened 2020-03-04
Support originating audience for skills conversations

### LINKED ISSUE #553: Parity issue with skill_conversation_id_factory
## Describe the bug
There is no concrete implementation of this in javascript or dotnet sdks.

Python:
https://github.com/microsoft/botbuilder-python/blob/master/libraries/botbuilder-core/botbuilder/core/skills/skill_conversation_id_factory.py

dotnet:
https://github.com/microsoft/botbuilder-dotnet/tree/master/libraries/Microsoft.Bot.Builder/Skills

https://github.com/microsoft/botbuilder-js/tree/master/libraries/botbuilder/src/skills

## Expected behavior
parity

[bug]

### LINKED ISSUE #798: Update SkillHandler to use ContinueConversationAsync overload with audience
## Describe the bug
SkillHandler needs to provide the correct audience when calling BotFrameworkAdapter.continue_conversation() (.net implementation: microsoft/botbuilder-dotnet#3409, waiting on https://github.com/microsoft/botbuilder-python/issues/780). It will get this audience from the SkillConversationIdFactory implementation.

### LINKED ISSUE #799: Update SkillHttpClient/BotFrameworkClient to support audience and bufferedReplies
## Describe the bug
The SkillHttpClient needs to include an overload that takes an audience and passes it through to the SkillConversationIdFactory for storage.

This audience is required so that when the SkillConversationIdFactory is used via the SkillHandler class, the correct ConnectorClient is created for sending activities.

We need another method to enable configuring `"bufferedReplies"` on `Activity.DeliveryMode` for forwarded activities.

## Additional context

waiting on https://github.com/microsoft/botbuilder-python/issues/780

PR https://github.com/microsoft/botbuilder-dotnet/pull/3409 that added ContinueConversationAsync overload to require audience.

### LINKED ISSUE #800: Update SkillConversationIdFactoryBase to take and return the correct Audience/OAuthScope
## Describe the bug
SkillConversationIdFactoryBase should have overloads or new methods that take and return audiences when storing and retrieving ConversationReferences.

This is required to light up the upcoming work in #780 (https://github.com/microsoft/botbuilder-dotnet/pull/3409/) and will ensure that proactive messages from bots to other bots is correctly setup.

@EricDahlvang is driving the initial implementation in .net (https://github.com/microsoft/botbuilder-dotnet/issues/3449) and has proposed additional changes

EricDahlvang:
Fixes #798
Fixes #799
Fixes #800
Fixes #553

`ConversationIdFactoryBase` is backwards compatible, and will work with legacy implementations as well as those which support `SkillConversationIdFactoryOptions`  (see attached .zip for a legacy implementation, and new implementations of `SkillConversationIdFactory `and `SkillHttpClient`)

Incidentally, I noticed the Python sdk does not have an implement of SkillHttpClient, while javascript and dotnet do.  Added a comment to https://github.com/microsoft/botbuilder-python/issues/553 regarding this.

[skill_factories_and_clients.zip](https://github.com/microsoft/botbuilder-python/files/4286035/skill_factories_and_clients.zip)


tracyboehrer:
1) I can't reliably attest that these are the appropriate Skills changes so I just compared against C#
2) Kudos on the ConversationIdFactoryBase method overload handling.  This is pretty much how Axel recommended handling this.  Use a Union or optional args and document in comments.
3) C# has the SkillConversationIdFactory in each bot (sample).  It's in BotBuilder.core in Python.  I don't understand the arguments for one vs the other.  Just a noted difference.  Though I see why you did it.
4) SkillConversationIdFactoryOptions: Axel has commented before on using @property instead of normal class member vars.  If your intention was to make them read-only (which doesn't really exist in Python since the caller can use the "protected" var) then OK.  But it doesn't appear that C# provides read-only members.   So I would just go with normal Python fields.  self.from_bot_auth_scope, et.c..
5) SkillConversationReference:  Same as #4

EricDahlvang:
Thank you @tracyboehrer !  I'll address these things, and tests, today.  Before merging this, we should also determine what we are going to do about the SkillConversationIdFactory and missing SkillHttpClient.  I'll bring this up in standup today.

PR #839, opened 2020-03-07
get member updates

### LINKED ISSUE #809: [PORT] Get member changes for Teams
https://github.com/microsoft/botbuilder-dotnet/pull/3473

Adding roster change with no alterations to the IConversations interface. Calling getMember would require a downcast from an object of type IConversations. See line 78 of TeamsInfo.cs for an example of what this looks like.

This PR includes:
- Adding support for the get single member endpoint to the Conversations object
- Adding support for get single member endpoint to the TeamsInfo object
- Adding support for getting paged members to the TeamsInfo object

Also fixed the formatting in ConversationsExtensions, because reasons.

Resolves #3447 

# Changed projects
* Microsoft.Bot.Builder
* Microsoft.Bot.Connector
* Microsoft.Bot.Schema

clearab:
fixes #809 

Add
 - support for get member by memberId
- get member by Id for Teams
- get paged members for Teams


libraries/botbuilder-core/botbuilder/core/bot_framework_adapter.py
```diff
@@ -770,6 +770,44 @@ def _create_context(self, activity):
...
+            if not member_id:
+                raise TypeError(
+                    "BotFrameworkAdapter.get_conversation_member(): missing memberId"
+                )
+
+            client = context.turn_state[BotAdapter.BOT_CONNECTOR_CLIENT_KEY]
+            return await client.conversations.get_conversation_member(
+                context.activity.conversation.id, member_id
+            )
+        except Exception as error:
```
EricDahlvang: I don't think this try/catch is helpful, since it is just raising the same error.

libraries/botbuilder-core/botbuilder/core/channel_service_handler.py
```diff
@@ -343,6 +351,24 @@ def __init__(
         """
         raise BotActionNotImplementedError()
 
+    async def on_get_conversation_member(
+        self, claims_identity: ClaimsIdentity, conversation_id: str, member_id: str,
+    ) -> ChannelAccount:
+        """
+        get_conversation_member() API for Skill.
```
EricDahlvang: 'for Skill'?

libraries/botbuilder-core/botbuilder/core/teams/teams_info.py
```diff
@@ -103,6 +106,78 @@ class TeamsInfo:
...
+            raise TypeError("TeamsInfo.get_team_member: method requires a member_id")
+
+        connector_client = await TeamsInfo._get_connector_client(turn_context)
+        return await TeamsInfo._get_member(
+            connector_client, turn_context.activity.conversation.id, member_id
+        )
+
+    @staticmethod
+    async def get_member(
+        turn_context: TurnContext, member_id: str
```
EricDahlvang: Should there be a None check on member_id here?

tracyboehrer: I would think not since the methods this calls with member_id do check.

libraries/botbuilder-core/botbuilder/core/channel_service_handler.py
clearab: Copy pasta from get_conversation_members(), actually have no idea why it is there

libraries/botbuilder-core/botbuilder/core/bot_framework_adapter.py
tracyboehrer: I would agree it's not needed

libraries/botbuilder-core/botbuilder/core/channel_service_handler.py
EricDahlvang: looked at this some more... since it is in channel_service_handler, it IS part of the api for a skill.

PR #860, opened 2020-03-12
Skills: Cancel dialogs only if EOC is coming from parent.

### LINKED ISSUE #859: [PORT] EoC should only be handled when coming from parent in RunAsync
https://github.com/microsoft/botbuilder-dotnet/pull/3540

Fixes #3541 

Added check to only cancel dialogs in RunAsync when EoC comes from a parent bot (root or skill).

The issue was that this affected part of the code was being triggered when a skill responds with an EoC to a caller (EoCs are routed back to the pipeline in SkillHandler). 

This PR also applies the urn format to CallerId as per [OBI Spec](https://github.com/microsoft/botframework-obi/blob/master/protocols/botframework-activity/botframework-activity.md#bot-calling-skill)

# Changed projects
* Microsoft.Bot.Builder.Dialogs
* integration

tracyboehrer:
Fixes #859

### DIFF
--- a/libraries/botbuilder-dialogs/botbuilder/dialogs/dialog_extensions.py
+++ b/libraries/botbuilder-dialogs/botbuilder/dialogs/dialog_extensions.py
@@ -30,7 +30,13 @@ class DialogExtensions:
             if (
                 turn_context.activity.type == ActivityTypes.end_of_conversation
                 and dialog_context.stack
+                and DialogExtensions.__is_eoc_coming_from_parent(turn_context)
             ):
+                remote_cancel_text = "Skill was canceled through an EndOfConversation activity from the parent."
+                await turn_context.send_trace_activity(
+                    f"Extension {Dialog.__name__}.run_dialog", label=remote_cancel_text,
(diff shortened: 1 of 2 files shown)


libraries/botbuilder-dialogs/botbuilder/dialogs/dialog_extensions.py
```diff
@@ -75,3 +81,9 @@ class DialogExtensions:
             results = await dialog_context.continue_dialog()
             if results.status == DialogTurnStatus.Empty:
                 await dialog_context.begin_dialog(dialog.id)
+
+    @staticmethod
+    def __is_eoc_coming_from_parent(turn_context: TurnContext) -> bool:
+        # To determine the direction we check callerId property which is set to the parent bot
+        # by the BotFrameworkHttpClient on outgoing requests.
+        return not turn_context.activity.caller_id
```
axelsrz: I believe the return value should be:
`return bool(turn_context.activity.caller_id)`

Issue #974, opened 2020-04-21
Change onTeamsMemberAdded for Teams to use getMember() (Python)


clearab:
See [parent](https://github.com/microsoft/botframework-sdk/issues/5810)

PR #1009, opened 2020-04-28
Change onTeamsMemberAdded for Teams to use getMember()

EricDahlvang:
Fixes: https://github.com/microsoft/botbuilder-python/issues/974


EricDahlvang:
This repo needs more tests for teams_activity_handler and teams_info.

Please let me know if the test i did add is an acceptable way to do this, and I'll add some more.

PR #1038, opened 2020-05-01
Support SSO with skill dialog and expected replies

### LINKED ISSUE #918: Add support to SkillDialog so SSO works with expectReply (Python)
See [parent](https://github.com/microsoft/botframework-sdk/issues/5780)

tracyboehrer:
Fixes #918 

Port of https://github.com/microsoft/botbuilder-dotnet/pull/3818


libraries/botbuilder-core/botbuilder/core/adapters/test_adapter.py
```diff
@@ -446,6 +447,23 @@ def add_exchangeable_token(
...
+    ):
+        key = ExchangeableToken(
+            connection_name=connection_name,
+            channel_id=channel_id,
+            user_id=user_id,
+            exchangeable_item=exchangeable_item,
+            token=TestAdapter.__EXCEPTION_EXPECTED,
+        )
+
+        self.exchangeable_tokens[key.to_key()] = key
```
stevengum, line 465: I noticed the same code in JavaScript; in JS you can use objects as keys [Maps](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map), so I might change the JS impl to do that (parity with C#). 

I don't recall, but I believe there is something similar in Python...

Note, this isn't actionable for this PR, just potential future code cleanup.

libraries/botbuilder-schema/botbuilder/schema/_models_py3.py
```diff
@@ -287,6 +288,28 @@ def __init__(
...
         self.caller_id = caller_id
 
+    def create_reply(self, text: str = None, locale: str = None):
+        return Activity(
+            type="message",
+            timestamp=datetime.utcnow(),
+            from_property=ChannelAccount(
+                id=self.recipient.id if self.recipient else None,
+                name=self.recipient.name if self.recipient else None,
+            ),
```
EricDahlvang, line 298:
```suggestion
            ),
            recipient = ChannelAccount(
                id=self.from_property.id if self.from_property else None,
                name=self.from_property.name if self.from_property else None,
            ),
```

ref: https://github.com/microsoft/botbuilder-dotnet/blob/master/libraries/Microsoft.Bot.Schema/ActivityEx.cs#L170

libraries/botbuilder-dialogs/botbuilder/dialogs/skills/skill_dialog.py
```diff
@@ -226,8 +228,86 @@ async def _send_to_skill(
...
+            self.dialog_options.bot_id,
+            skill_info.app_id,
+            skill_info.skill_endpoint,
+            self.dialog_options.skill_host_endpoint,
+            incoming_activity.conversation.id,
+            activity,
+        )
+
+        # Check response status: true if success, false if failure
+        return response.status / 100 == 2
```
carlosscastro, line 313: First time I see this, interesting. Is there no built in or popular function in python for this? i.e. dotnet has HttpStatusCodes and IsSuccessStatusCode() in a package that gets added everywhere anyways. If it requires adding a new package for this check, then it is not worth, but just checking.

libraries/botbuilder-dialogs/tests/test_skill_dialog.py
```diff
@@ -217,8 +220,223 @@ async def test_should_throw_on_post_failure(self):
...
+                SkillDialogTests.create_oauth_card_attachment_activity("https://test")
+            ]
+        )
+
+        sequence = 0
+
+        async def post_return():
+            nonlocal sequence
+            if sequence == 0:
+                result = InvokeResponse(body=first_response, status=200)
```
carlosscastro, line 236: In C# and js we have constants for status. HttpStatusCodes. Thoughts around adopting something similar? Have you considered this already?

tracyboehrer: There is.  You don't like magic numbers, Carlos? ;-)

PR #1051, opened 2020-05-05
Updated master to 4.10.0

tracyboehrer:


libraries/botbuilder-adapters-slack/botbuilder/adapters/slack/about.py
```diff
@@ -5,7 +5,7 @@
 
 __title__ = "botbuilder-adapters-slack"
 __version__ = (
-    os.environ["packageVersion"] if "packageVersion" in os.environ else "4.7.1"
+    os.environ["packageVersion"] if "packageVersion" in os.environ else "4.10"
```
axelsrz, line 8: I believe this should be 4.10.0

This applies to several instances.

PR #1067, opened 2020-05-14
Do NOT call TeamsInfo.get_member for the bot: cherry pick 4.9 fix (#1066)

### LINKED ISSUE #1065: onTeamsMembersAddedEvent handler doesn't run when bot is the member
When a bot receives an onTeamsMemberAdded event where the member added was the bot, a 502 error is thrown, which results in this event never being fired.

Cause:
The event calls the getMember API, which is the source of the 502. This API cannot be passed a bot ID.

Fix:
Check to see if the member added is the bot, and if it is return a truncated TeamsChannelAccount.

EricDahlvang:
Fixes #1065 
* Do not call TeamsInfo.get_member for the bot

* fix botid for test


EricDahlvang:
Something is wrong with the status report here. :(  

any ideas @tracyboehrer ?

PR #1069, opened 2020-05-15
Adding TeamsInfo to on_teams_member_removed

Virtual-Josh:
on_teams_members_removed is not consistent with the JS and C#. This is a breaking change. I'm not sure how we want to handle this.


tracyboehrer:
@axelsrz @Virtual-Josh Axel, what is your opinion on this?  One option is to just induce the breaking change.  The other is to add the TeamsInfo argument to the end of the list and make it default to None.  While it would have a different signature than C#, it would at least allow the same argument to be supplied.  I lean toward to first option.

PR #1182, opened 2020-06-26
Add ActivityEx implementation

### LINKED ISSUE #1157: [Parity] ActivityEx is not present in Botbuilder-Python
See [dotnet source](https://github.com/microsoft/botbuilder-js/issues/1753)

denscollo:
Fixes #1157

## Description
Implements the methods from ActivityEx from DotNet in the Activity class. It also adds tests for the Activity class.

## Specific Changes
Added the Activity methods present in [ActivityEx from BotBuilder-DotNet](https://github.com/microsoft/botbuilder-dotnet/blob/master/libraries/Microsoft.Bot.Schema/ActivityEx.cs) to the `_models_py3.py` file.
Created the `test_activity.py` file and Added 55 tests for the methods in the Activity class.

## Testing
image


tracyboehrer:
@denscollo You'll need to update from master again, then merge.

denscollo:
Hi @tracyboehrer, we just updated the branch and it works fine! But we can't merge it since we don't have the permissions. Thanks!

PR #1240, opened 2020-07-10
return ResourceResponse

### LINKED ISSUE #1137: SkillHandler doesn't return ResourceResponse when forwarding activities (Python)
See [parent](https://github.com/microsoft/botframework-sdk/issues/5919)

axelsrz:
Fixes #1137


tracyboehrer:
@axelsrz Looks like we need to update the yaml for the new 3.6 and 3.7 versions.  Maybe there is a way to not specify the dot release so that we always use the latest and not regularly run into this?

axelsrz:
> @axelsrz Looks like we need to update the yaml for the new 3.6 and 3.7 versions. Maybe there is a way to not specify the dot release so that we always use the latest and not regularly run into this?

just submitted #1243

PR #1297, opened 2020-07-30
Add Buildin Function

ShYuPe:
Add buildin Function for adaptive expression.
Start from datetime.
Still working on it...


libraries/adaptive-expressions/adaptive/expressions/builtin_functions/add_to_time.py
```diff
@@ -0,0 +1,65 @@
...
+        if error is None:
+            time_format = (
+                args[3] if len(args) == 4 else FunctionUtils.default_date_time_format
+            )
+            if isinstance(args[1], numbers.Number) and isinstance(args[2], str):
+                value, error = AddToTime.eval_add_to_time(
+                    args[0], args[1], args[2], time_format
+                )
+                if time_format == FunctionUtils.default_date_time_format:
+                    value = value[:-4] + "Z"
```
cosmicshuai, line 30: is there a better way to do the formating?

ShYuPe: I tried to find a better way, but it seems that in python, a way like "fff" is not exist. And the function .strftime doesn't accept a format like "%.3f". So I think this is the only way to format the output.

PR #1310, opened 2020-08-04
Add "type checking" expression functions

Danieladu:
Add functions:
- [x] isBoolean
- [x] isString
- [x] isArray
- [x] isInteger
- [x] isFloat
- [x] isObject
- [ ] isDataTime

isDateTime depends on datetime functions that would be provided by @ShYuPe


libraries/adaptive-expressions/tests/test_expression_parser.py
```diff
@@ -1353,6 +1347,145 @@ def test_select(self):
         assert value == "zero3,one3,two3"
         assert error is None
 
+    def test_is_boolean(self):
```
feich-ms, line 1350: could you add tests to the bottom and add comments at head like # Type checking

PR #1326, opened 2020-08-06
Refactored SkillDialog to call ConversationFacotry.CreateConversationId only once

### LINKED ISSUE #1322: [PORT] Refactored SkillDialog to call ConversationFacotry.CreateConversationId only once
https://github.com/microsoft/botbuilder-dotnet/pull/4382

Refactored SkillDialog to call the factory only during begin and store the generated conversationId in state to be used in subsequent calls to the skill.

Needs parity for python and JS

Fixes #4374

# Changed projects
* Microsoft.Bot.Builder.Dialogs
* Microsoft.Bot.Builder.Dialogs.Tests

tracyboehrer:
Fixes #1322

### DIFF
--- a/libraries/botbuilder-dialogs/tests/test_skill_dialog.py
+++ b/libraries/botbuilder-dialogs/tests/test_skill_dialog.py
@@ -6,6 +6,7 @@ from typing import Callable, Union
 from unittest.mock import Mock
 
 import aiounittest
+from botframework.connector.token_api.models import TokenExchangeResource
 from botbuilder.core import (
     ConversationState,
     MemoryStorage,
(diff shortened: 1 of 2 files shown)


libraries/botbuilder-dialogs/tests/test_skill_dialog.py
```diff
@@ -148,10 +148,13 @@ async def capture(
             conversation_state=conversation_state,
         )
 
+        assert len(dialog_options.conversation_id_factory.conversation_refs) == 0
```
axelsrz, line 151: this is nitpicking but this could be changed to 
`assert not dialog_options.conversation_id_factory.conversation_refs`

Issue #1347, opened 2020-08-12
InstallationUpdate should have actions "add" and "remove"


johnataylor:
To fully model the protocol the InstallationUpdate should have two sub events corresponding to "add" and "remove"

These have been added to the JavaScript and C#.

PR #1368, opened 2020-08-26
Add installation update sub events to ActivityHandler

EricDahlvang:
Fixes https://github.com/microsoft/botbuilder-python/issues/1347

While adding this, I realized on_installation_update should be named on_installation_update_activity... yet, we have released it as on_installation_update in 4.10.0 (although this is not likely to be used by any bots just yet, since the event is not sent from Teams yet).  Any recommendation, should we update this event name to match the convention, or leave it?

### DIFF
--- a/libraries/botbuilder-core/botbuilder/core/activity_handler.py
+++ b/libraries/botbuilder-core/botbuilder/core/activity_handler.py
@@ -379,12 +379,12 @@ class ActivityHandler:
         :returns: A task that represents the work queued to execute
         """
         if turn_context.activity.action == "add":
-            return await self.on_installation_update_add_activity(turn_context)
+            return await self.on_installation_update_add(turn_context)
         if turn_context.activity.action == "remove":
-            return await self.on_installation_update_remove_activity(turn_context)
+            return await self.on_installation_update_remove(turn_context)
         return
 
-    async def on_installation_update_add_activity(  # pylint: disable=unused-argument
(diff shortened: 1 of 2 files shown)


johnataylor:
i can't really decide on the method name - i think we could leave it??? perhaps @tracyboehrer has an opinion.

tracyboehrer:
> i can't really decide on the method name - i think we could leave it??? perhaps @tracyboehrer has an opinion.

I guess I would say "_activity" seems unneeded (though I've always wondered that).  But specifically in this case, "on_installation_update" doesn't have that.  So to be consistent, I think "on_installation_update_add" and "on_installation_update_remove" maybe better.

But... what did we do on the other repos?

EricDahlvang:
Javascript has the 'activity' appended to the method name of all three:
https://github.com/microsoft/botbuilder-js/blob/main/libraries/botbuilder-core/src/activityHandler.ts#L576

csharp has the 'activity' appended only to the initial event: https://github.com/microsoft/botbuilder-dotnet/blob/main/libraries/Microsoft.Bot.Builder/ActivityHandler.cs#L560  but not the add and remove counterparts.  (Which is apparently the opposite of this python PR, just to confuse things further).

EricDahlvang:
@tracyboehrer I've removed '_activity' from the add and remove method names.

tracyboehrer (approved):

PR #1379, opened 2020-09-04
Cherry-Pick Skill Validation to 4.7 (#1375)

EricDahlvang:


axelsrz (approved):

axelsrz:
This was force pushed due that only fails in 3.8 (not specifically supported at that time) and fails on tests that has nothing to do with this change.

PR #1386, opened 2020-09-17
Include API results for LuisRecognizerOptionsV3 to work with np-with-…

parmarjigar:
…dispatch example to get API response

Fixes #
Unlike V2, when working with nlp-with-dispatch with V3 it throwed error of not having the key luisResult in recognizer_result.properties. Though we passed include_api_results=True, it didnt have any support to include the API response for V3

## Description

I added the code that if include_api_results=True, it will store the api response as it did for V2

## Specific Changes

  - 
  -
  -


cleemullins:
@munozemilio @axelsrz We need a plan here. This PR is 9 months old, and is now full-term.

PR #1399, opened 2020-10-05
Avoid typing activity if is skill

### LINKED ISSUE #1388: Avoid sending typing activity when bot is invoked as skill
We should port this once the C# PR is merged.
See [parent](https://github.com/microsoft/botframework-sdk/issues/6049)

axelsrz:
Fixes #1388


gabog (approved):
Looks good to me but would wait for @tracyboehrer to make sure the python is good.

PR #1417, opened 2020-10-21
Teams CacheInfo

### LINKED ISSUE #1408: Add support for CacheInfo in Bot invoke response (Python)
See [parent](https://github.com/microsoft/botframework-sdk/issues/6006)

tracyboehrer:
Fixes #1408

### DIFF
--- a/libraries/botbuilder-schema/botbuilder/schema/teams/_models_py3.py
+++ b/libraries/botbuilder-schema/botbuilder/schema/teams/_models_py3.py
@@ -735,7 +735,12 @@ class MessagingExtensionActionResponse(Model):
     }
 
     def __init__(
-        self, *, task=None, compose_extension=None, cache_info: CacheInfo=None, **kwargs
+        self,
+        *,
+        task=None,
+        compose_extension=None,
+        cache_info: CacheInfo = None,
+        **kwargs
     ) -> None:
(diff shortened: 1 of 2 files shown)


libraries/botbuilder-schema/botbuilder/schema/teams/_models_py3.py
```diff
@@ -707,12 +731,16 @@ class MessagingExtensionActionResponse(Model):
             "key": "composeExtension",
             "type": "MessagingExtensionResult",
         },
+        "cache_info": {"key": "cacheInfo", "type": "CacheInfo"},
     }
 
-    def __init__(self, *, task=None, compose_extension=None, **kwargs) -> None:
+    def __init__(
+        self, *, task=None, compose_extension=None, cache_info: CacheInfo, **kwargs
```
axelsrz, line 738: shouldn't the new parameter be optional? And if sure that not (would be a breaking change) it should be placed before optional parameters

axelsrz (changes requested):

libraries/botbuilder-schema/botbuilder/schema/teams/_models_py3.py
EricDahlvang: definitely should be optional

axelsrz (approved):

PR #1418, opened 2020-10-22
Add on_teams_team_renamed "overload"

### LINKED ISSUE #1405: Teams added, deleted and renamed API quite confusing?
## Version
4.10.1

## Describe the bug
While testing the application  on_teams_members_added , on_teams_members_removed is working fine but on_teams_team_renamed is not working instead of we have to use on_teams_team_renamed_activity api. 

**Note** : channel added , renamed , deleted events is working fine.

## Expected behavior
on_teams_team_renamed api should be work.

mdrichardson:
Fixes #1405 

## Description

`on_teams_team_renamed_activity()` was mistakenly named against the pattern of the rest of the activity handlers (which don't end in "activity"`

## Specific Changes

- Add overload method called `on_teams_team_renamed()`
- Add deprecation notice for `on_teams_team_renamed_activity()`

### DIFF
--- a/libraries/botbuilder-core/botbuilder/core/teams/teams_activity_handler.py
+++ b/libraries/botbuilder-core/botbuilder/core/teams/teams_activity_handler.py
@@ -529,7 +529,7 @@ class TeamsActivityHandler(ActivityHandler):
                         channel_data.channel, channel_data.team, turn_context
                     )
                 if channel_data.event_type == "teamRenamed":
-                    return await self.on_teams_team_renamed_activity(
+                    return await self.on_teams_team_renamed(
                         channel_data.team, turn_context
                     )
                 if channel_data.event_type == "teamRestored":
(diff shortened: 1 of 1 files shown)


libraries/botbuilder-core/botbuilder/core/teams/teams_activity_handler.py
```diff
@@ -529,7 +529,7 @@ async def on_conversation_update_activity(self, turn_context: TurnContext):
                         channel_data.channel, channel_data.team, turn_context
                     )
                 if channel_data.event_type == "teamRenamed":
-                    return await self.on_teams_team_renamed_activity(
+                    return await self.on_teams_team_renamed(
```
mdrichardson, line 532: This was actually a little trickier to implement than just using a simple overload because of this line. One of `on_teams_team_renamed_activity()` and on_teams_team_renamed()` *has* to call the other (line 615). I tried to consider how this impacts backwards-compat (below), but please look it over and think it through, letting me know if I missed anything.

**Regarding backwards-compat, there's really two ways to do this and I'm open to changing this**

### 1. (current PR) Have `on_conversation_update_activity()` call `on_teams_team_renamed()`

The flow works like this:

1. Activity comes in to `on_conversation_update_activity()`
2. `on_conversation_update_activity()` calls `on_teams_team_renamed()`
3. `on_teams_team_renamed()` calls `on_teams_team_renamed_activity()`, if it hasn't been overridden

**Works for developers who:**

- Have already overridden `on_conversation_update_activity()`
- Have already overridden `on_teams_team_renamed_activity()`
- Have already overridden both `on_conversation_update_activity()` and `on_teams_team_renamed_activity()`

**Does NOT work for developers who:**

- Have created a method `on_teams_team_renamed()` AND overridden `on_teams_team_renamed_activity()`, since `on_conversation_update_activity()` will no longer call `on_teams_team_renamed_activity()`
    - This is pretty unlikely since creating `on_teams_team_renamed()` would have accomplished nothing.

### 1. (alternative PR) Have `on_conversation_update_activity()` call `on_teams_team_renamed_**activity**()`

The difference is basically that there are no changes made to `on_conversation_update_activity()` and `on_teams_team_renamed_activity()` calls `on_teams_team_renamed()`

The flow would work like this:

1. Activity comes in to `on_conversation_update_activity()`
2. `on_conversation_update_activity()` calls `on_teams_team_renamed_**activity**()`
3. `on_teams_team_renamed_activity()` calls `on_teams_team_renamed()`, if it hasn't been overridden

**Works for developers who:**

- Have already overridden `on_conversation_update_activity()`
- Have already overridden `on_teams_team_renamed_activity()`
- Have already overridden both `on_conversation_update_activity()` and `on_teams_team_renamed_activity()`

**Does NOT work for developers who:**

- Works for all since if the developer has created `on_teams_team_renamed()`, they're already going to use it how they please.

## Summary

I went with #1 because it keeps `on_conversation_update_activity()` aligned through all the methods so that if future developers override it, it makes more "sense". However, option #1 also has potential backwards-compat issues, albeit very minimal.

axelsrz (approved):

PR #1459, opened 2020-12-29
fix type annotation

asdkant-ey:
Fixes #
#1458

## Description
VSCode's pylance complained about this, the message was:
```
List expression not allowed in type annotation
  Use List[T] instead
```
It'd be a good idea to scan all the code for other occurrences of the issue.

## Specific Changes
- Changed `[str]` to `List[str]` in the type annotation of `get_suggestions_card()`

### DIFF
--- a/libraries/botbuilder-ai/botbuilder/ai/qna/utils/qna_card_builder.py
+++ b/libraries/botbuilder-ai/botbuilder/ai/qna/utils/qna_card_builder.py
@@ -14,7 +14,7 @@ class QnACardBuilder:
 
     @staticmethod
     def get_suggestions_card(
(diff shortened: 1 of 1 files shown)


tracyboehrer (approved):

tracyboehrer (changes requested):
This does produce an error during build:

libraries/botbuilder-ai/botbuilder/ai/qna/utils/qna_card_builder.py:17: in QnACardBuilder
    suggestions: List[str], card_title: str, card_no_match: str
E   NameError: name 'List' is not defined

srinaath:
@tracyboehrer Can we get some traction on this PR from the customer.

tracyboehrer:
Replaced by #1491

PR #1486, opened 2021-02-01
Adaptive Card Tabs

### LINKED ISSUE #1463: Adaptive Card Tabs (Python)
See [parent](https://github.com/microsoft/botframework-sdk/issues/6138)

axelsrz:
Fixes #1463


libraries/botbuilder-core/botbuilder/core/teams/teams_activity_handler.py
```diff
@@ -835,3 +853,29 @@ async def on_teams_channel_restored(  # pylint: disable=unused-argument
...
+    ):
+        """
+        Override this in a derived class to provide logic for when a tab is fetched.
+
+        :param turn_context: A context object for this turn.
+        :param tab_request: The tab invoke request value payload.
+
+        :returns: A Tab Response for the request.
+        """
+        return
```
mdrichardson, line 868: I *believe* these should return a NOT_IMPLEMENTED invoke response, similar to [`on_teams_messaging_extension_fetch_task()`](https://github.com/microsoft/botbuilder-python/blob/f53312fb4344ee9501face7582c7f5e2b799a402/libraries/botbuilder-core/botbuilder/core/teams/teams_activity_handler.py#L420). [They do in .NET](https://github.com/microsoft/botbuilder-dotnet/blob/0fde312e927d7a5f651969ed133773eb38067b64/libraries/Microsoft.Bot.Builder/Teams/TeamsActivityHandler.cs#L400) and I think it's only the `OnTeamDeleted/Added/etc` that don't.

libraries/botbuilder-schema/botbuilder/schema/teams/_models_py3.py
```diff
@@ -2058,3 +2086,265 @@ def __init__(
...
+     ~botframework.connector.teams.models.TabEntityContext
+    :param context: Gets or sets current tab entity request context.
+    :type context:
+     ~botframework.connector.teams.models.TabContext
+    :param state: Gets or sets state, which is the magic code for OAuth Flow.
+    :type state: str
+    """
+
+    _attribute_map = {
+        "tab_entity_context": {"key": "tabEntityContext", "type": "TabEntityContext"},
```
mdrichardson, line 2127: Will this look for the JSON property, `tabContext`? See [.NET](https://github.com/microsoft/botbuilder-dotnet/blob/cb33e75b72735e25177fb82084813dcaeadf16c7/libraries/Microsoft.Bot.Schema/Teams/TabRequest.cs#L27)

libraries/botbuilder-core/botbuilder/core/teams/teams_activity_handler.py
axelsrz: Thanks a lot for piointing this! I moved the methods to their corresponding place on added this fix

libraries/botbuilder-schema/botbuilder/schema/teams/_models_py3.py
axelsrz: Fixed

libraries/botbuilder-schema/botbuilder/schema/teams/_models_py3.py
```diff
@@ -2058,3 +2086,265 @@ def __init__(
...
+class TabResponseCards(Model):
+    """
+    Envelope for cards for a TabResponse.
+
+    :param cards: Gets or sets adaptive card for this card tab response.
+    :type cards: list[object]
+    """
+
+    _attribute_map = {
+        "cards": {"key": "cards", "type": "[object]"},
```
mdrichardson, line 2175: @axelsrz Are you able to change the type to something like `[TabResponseCard]` or whatever the Python equivalent is of [this](https://github.com/microsoft/botbuilder-js/pull/3224/files#diff-348dcf11b9d343d90e766394f8ebae3c3675bde51e95705520bd15451f0c5720R1308) (an array of `TabResponseCard`).

@EricDahlvang Do we still need this change in .NET?

libraries/botbuilder-schema/botbuilder/schema/teams/_models_py3.py
```diff
@@ -2058,3 +2086,265 @@ def __init__(
...
+    :param context: Gets or sets current tab entity request context.
+    :type context:
+     ~botframework.connector.teams.models.TabContext
+    :param data: User input data. Free payload containing properties of key-value pairs.
+    :type data:
+     ~botframework.connector.teams.models.TabSubmitData
+    """
+
+    _attribute_map = {
+        "tab_entity_context": {"key": "tabEntityContext", "type": "TabEntityContext"},
```
mdrichardson, line 2236: I believe we need [the same `tabContext` change](https://github.com/microsoft/botbuilder-dotnet/blob/cb33e75b72735e25177fb82084813dcaeadf16c7/libraries/Microsoft.Bot.Schema/Teams/TabSubmit.cs#L27) here.

libraries/botbuilder-schema/botbuilder/schema/teams/_models_py3.py
```diff
@@ -2058,3 +2086,265 @@ def __init__(
...
+    :param type: Currently, 'tab/submit'.
+    :type type: str
+    :param properties: Gets or sets properties that are not otherwise defined by the TabSubmit
+     type but that might appear in the serialized REST JSON object.
+    :type properties: object
+    """
+
+    _attribute_map = {
+        "type": {"key": "type", "type": "str"},
+        "properties": {"key": "properties", "type": "object"},
```
mdrichardson, line 2267: I don't believe this will parse correctly. The [spec](https://github.com/microsoft/botframework-sdk/issues/6138) is:

```json
{
    "name": "tab/submit",
    "value": {
        "data": {
            "type": "tab/submit",
            // ... <data properties>
        },
        [...]
}
```

My understanding is that there won't be a `properties` key, so much as an unknown amount of un-named keys. @EricDahlvang @clearab Is this correct?

If so, [here's how I handled it in JS](https://github.com/microsoft/botbuilder-js/pull/3224/files#diff-348dcf11b9d343d90e766394f8ebae3c3675bde51e95705520bd15451f0c5720R1365) (not sure what the Py equivalent is).

EricDahlvang: This should be handled similarly to the Properties field on Entity, ConversationAccount, ChannelAccount, Activity, etc.

EricDahlvang:
@axelsrz i think this looks good, but need a sample to test against... can you port this one: https://github.com/microsoft/BotBuilder-Samples/pull/3072

libraries/botbuilder-schema/botbuilder/schema/teams/_models_py3.py
EricDahlvang: @mdrichardson this if fixed in dotnet in this PR: https://github.com/microsoft/botbuilder-dotnet/pull/5185/files#diff-5a0a768d82736568f8e996abff4ff3eacea0696d1ed47f928822bfc2d6f895d2R30

EricDahlvang (approved):
Thanks @axelsrz !

PR #1515, opened 2021-02-10
Update azure blob SDK version

mdrichardson:
Partial fix for #1343 

# Please Read

#1343 is for porting over Azure Blobs and Queues from .NET and JS. In those SDK, Azure Blobs was plural-ified from `Bot.Builder.Azure.BlobStorage` to `...Blob*s*Storage`, adding a new Blob Storage library. This was done because of a dependency namespace conflict that would break binary compatibility.

As far as I can tell, we can't do this in the Python SDK because that would require the SDK to depend on two different versions of [`azure.storage.blob`](https://pypi.org/project/azure-storage-blob/). The current version in our SDK is 2.1, but the most recent version is 12.7.1. This PR updates to 12.7.0 because there's an msrest dependency conflict at 122.7.1.

So, this PR updates our current `blob_storage` to just use the new SDK. If approved, I'll then add an additional `queue_storage`.

There's some pretty big changes in the the internal code for this PR due to:

1. Forced changes when upgrading `azure.storage.blob` 10 major versions, and
2. Trying to make the code more or less match [.NET's BlobsStorage](https://github.com/microsoft/botbuilder-dotnet/blob/main/libraries/Microsoft.Bot.Builder.Azure.Blobs/BlobsStorage.cs) and [Node's BlobsStorage](https://github.com/microsoft/botbuilder-js/blob/main/libraries/botbuilder-azure-blobs/src/blobsStorage.ts)

## Testing

All tests pass when running Storage Emulator and setting [this to True](https://github.com/mdrichardson/botbuilder-python/blob/azureUpgrade/libraries/botbuilder-azure/tests/test_blob_storage.py#L23) so the tests run.

image

The warning are all of this nature:

image
image

...but per @axelsrz, that's expected for our SDK and Python 3.8.

### Backwards Compatibility Testing

I also tested by:

1. Downloading State Management sample and converting to use original blob_storage
2. Conversed with bot
3. Removed `botbuilder.azure` from sample, then added new `blob_storage.py` and ensured it used that
4. Started a conversation with the bot as the same user
5. Bot was able to retrieve previous conversation data from older `blob_storage`

image

Conversation and UserState data still look the same:

image

I'm not sure why e_tag gets quotes placed around it. I don't do any encoding, so I think it's something done by `azure.storage.blob`. I sort of mitigate it [here](https://github.com/microsoft/botbuilder-python/pull/1515/files#diff-d6a2674ca3d1f83710037d78c45d954c66d8e0e72a50948449d70885a9c1ca44R197)

#### Note
This library will eventually benefit from [investigating some of the Blob Storage options](https://github.com/microsoft/botbuilder-dotnet/issues/4684), but I won't have the bandwidth to land this in R12.


libraries/botbuilder-azure/botbuilder/azure/blob_storage.py
```diff
@@ -23,56 +44,107 @@ def __init__(
...
+        :return:
+        """
         if changes is None:
             raise Exception("Changes are required when writing")
         if not changes:
             return
 
-        self.client.create_container(self.settings.container_name)
-        self.client.set_container_acl(
-            self.settings.container_name, public_access=PublicAccess.Container
```
mdrichardson, line 72: I didn't see this referenced in .NET or JS and tests all passed fine without it. I can add it back, if necessary.

libraries/botbuilder-azure/botbuilder/azure/blob_storage.py
```diff
@@ -81,39 +153,54 @@ async def write(self, changes: Dict[str, object]):
...
-                    if_match=e_tag,
+
+            if e_tag:
+                await blob_reference.upload_blob(
+                    item_str, match_condition=MatchConditions.IfNotModified, etag=e_tag
                 )
-            except Exception as error:
-                raise error
+            else:
+                await blob_reference.upload_blob(item_str, overwrite=True)
```
mdrichardson, line 164: Match conditions and eTags changed quite a bit and I'm not 100% confident that I implemented this correctly. I tried some other approaches and this was the one that made all of our tests pass. Please review this part closely.

mdrichardson:
Converting back to draft. Looks like the `azure-storage-blob` package adds some quotes to some of the blob references, which breaks backwards-compat. I partially handled this [here](https://github.com/microsoft/botbuilder-python/pull/1515/files#diff-d6a2674ca3d1f83710037d78c45d954c66d8e0e72a50948449d70885a9c1ca44R197), but am fixing elsewhere.

axelsrz (approved):

PR #1559, opened 2021-03-04
Updates to slack-adapter

### LINKED ISSUE #1548: Received Slack payloads are not processed properly
## Version
v4.11.0

## Describe the bug
When content received from Slack is in type _application/x-www-form-urlencoded_, a dictionary called _request_dict_ is created from the body using the _query_string_to_dictionary_ method of the _SlackHelper_ class. The resulting dictionary has one key called _payload_ and its value is a dictionary with several key-values ({"type": ..., "user": ..., etc.}).
Later, a _SlackPayload_ object is created from the key-values. However, the constructor of the _SlackPayload_ class gets the values of the nested dictionary without getting the value of first key. Therefore, everything turns to be None in that object.

## To Reproduce
1. Use the _60.slack-adapter_ sample.
2. Put a breakpoint in line 11 of botbuilder/adapters/slack/slack_payload.py and start debugging.
3. Use the /test command in Slack (requires registering the command in the Slack application).
4. In Slack, click on of the buttons sent from the bot.

## Expected behavior
The created _SlackPayload_ object should contain information about the Slack action.

## Screenshots
ms_bot_bug

tracyboehrer:
Fixes #1548

There have been some some updates to the DotNet slack-adapter, and this PR includes most of those changes.


omid-jf:
I noticed in my tests that channel_id of a ConversationReference object (which is always "slack") is not the same as the channel ID that chat.delete API requires.
Therefore, I believe in line 137 of slack_adapter.py in this PR, _channel=reference.channel_id_ should be replaced by _channel=reference.conversation.id_
Moreover, since the current code of slack_helper.py does not set the timestamp property of activities, exception on line 134 of slack_adapter.py in this PR is always raised. There is a fix for this in PR #1605 (a conversion from datetime to epoch is required if using that PR) or _ts=context.activity.timestamp_ in line 137 of slack_adapter.py in this PR can be replaced by _ts=reference.activity_id_ since we already set activity_id to be the timestamp.

tracyboehrer:
@omid-jf Thanks!  I'll take a look and apply the fixes.

libraries/botbuilder-adapters-slack/botbuilder/adapters/slack/slack_adapter.py
```diff
@@ -32,10 +33,12 @@ def __init__(
         self,
         client: SlackClient,
         on_turn_error: Callable[[TurnContext, Exception], Awaitable] = None,
+        options: SlackAdapterOptions = None,
     ):
         super().__init__(on_turn_error)
         self.slack_client = client
         self.slack_logged_in = False
+        self.options = options if options else SlackAdapterOptions()
```
axelsrz, line 41: very small pet peeve but this could be changed to:
`self.options = options or SlackAdapterOptions`

axelsrz (approved):
Approved with comments

PR #1579, opened 2021-03-19
Only show "NO MATCH" response button if active learning is active

### LINKED ISSUE #1571: "no match response" button present in qna maker dialog when active learning is disabled
Python tracking issue for repo code-owners

See original issue for details: microsoft/botframework-sdk#6146

Zerryth:
Fixes #1571 

## Description
Currently when using the multi-turn feature in QnAMakerDialog, it will show "None of the above" 

*Multi-turn prompt has a 3rd extra button, "None of the above"* (Incorrect behavior)
image

"None of the above" is the "NO MATCH" response for active learning when QnAMaker wants explicit feedback about which of the closely scored answers the user considers is the actual answer their question correctly (in "NO MATCH" case, it means that none of the suggestions that are offered answers the user's question)

This "NO MATCH" response is only supposed to show up in an active learning prompt, however it is incorrectly also showing up fo rmulti-turn as well. See below

## Testing
1. Ensure that multi-turn prompt **does not** have NO MATCH response present
    -  image

2. Ensure that NO MATCH response **does** show up for active learning prompts:
    -  image

### DIFF
--- a/libraries/botbuilder-ai/botbuilder/ai/qna/utils/qna_card_builder.py
+++ b/libraries/botbuilder-ai/botbuilder/ai/qna/utils/qna_card_builder.py
@@ -67,13 +67,6 @@ class QnACardBuilder:
             for prompt in result.context.prompts
         ]
 
-        # Add No match text
-        button_list.append(
-            CardAction(
-                value=card_no_match_text, type="imBack", title=card_no_match_text,
-            )
-        )
-
         attachment = CardFactory.hero_card(HeroCard(buttons=button_list))
 
         return Activity(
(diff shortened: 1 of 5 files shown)


libraries/botbuilder-ai/botbuilder/ai/qna/utils/qna_card_builder.py
```diff
@@ -67,13 +67,6 @@ def get_qna_prompts_card(result: QueryResult, card_no_match_text: str) -> Activi
             for prompt in result.context.prompts
         ]
 
-        # Add No match text
```
Zerryth, line 70: - `get_qna_prompts_card` gets called for multi-turn qna dialog scenarios
- It's not an active learning scenario
- remove active learning NO MATCH appending from `get_qna_prompts_card`
- the NO MATCH response is still present within `get_suggestions_card`, which is what is called for active learning prompt scenarios

Zerryth:
@tracyboehrer @axelsrz could someone please review this PR? Want to get this in before code complete date

axelsrz (approved):

PR #1671, opened 2021-05-07
[#1670][Dialogs] Remove hardcoded InputHint for AttachmentPrompt

### LINKED ISSUE #1670: [Dialogs] InputHint is hardcoded for AttachmentPrompt as expecting_input
## Version
4.12.0

## Describe the bug
In the [AttachmentPrompt](https://github.com/microsoft/botbuilder-python/blame/main/libraries/botbuilder-dialogs/botbuilder/dialogs/prompts/attachment_prompt.py) class, the **prompt.input_hint** is set as [expecting_input](https://github.com/microsoft/botbuilder-python/blame/main/libraries/botbuilder-dialogs/botbuilder/dialogs/prompts/attachment_prompt.py#L45) replacing any value the user could set.
This behavior is not present in either .NET or JS SDKs.

## To Reproduce
1. Using the sample: [05.multi-turn-prompt](https://github.com/microsoft/BotBuilder-Samples/tree/main/samples/python/05.multi-turn-prompt):
    - Open [dialogs/user_profile_dialog.py](https://github.com/microsoft/BotBuilder-Samples/blob/main/samples/python/05.multi-turn-prompt/dialogs/user_profile_dialog.py) file.
    - Add the following line to the imports section `from botbuilder.schema import InputHints`.
    - Update picture_step method adding `input_hint=InputHints.accepting_input` to the message as can be seen below:
image
    - Save the changes.
    - Open the console in the bots folder and run the following commands:
        -  python -m pip install -r requirements.txt
        - python app.py

2. Open BotFramework Emulator and connect to your bot.
3. Follow the dialog flow until the profile picture step.
4. See the activity in the Inspection panel and check the value of the inputHint property.
image

## Expected behavior
The bot should send the activity with the inputHint property that was set for the message in the prompt options.

denscollo:
#minor -> replace for fixes x when promoted
Fixes #1670

## Description
This PR fixes the issue with the attachment prompt forcing the Input Hint on prompt and adds a new unit test to validate the change.

## Specific Changes

  - Remove lines in `attachment_prompt.py` that force the input_hint.
  - Add unit test `test_attachment_prompt_with_input_hint` that uses an activity with the input hint set, and validates that when used the properties are not changed.

## Testing

Below is the added unit test working before and after the change is applied.
image

And below is the 05.multi-turn botbuilder sample using the library with the changes applied.
image


tracyboehrer (approved):

tracyboehrer:
@denscollo  or @santgr11 Can you update from main again and I'll get this merged.

PR #1881, opened 2022-01-05
Avoid exceptions for request without Content-Type header

ivopisarovic:
## Description
If a non-bot request comes to the server, it raised an exception because some request may not have content-type header.

## Specific Changes
  - Improved the condition to check if the content-type header is present before checking its value.

## Testing
  - Nothing new added.


tracyboehrer:
We can't make changes to previous release branches.  Is this still applicable to main?

ivopisarovic:
I’m afk now. I will look at it in 2 weeks.

ivopisarovic:
@tracyboehrer I looked at it an the affected files has not been changed since then. Hence, the same change applies also to main. I am closing this and opening a new PR into main: https://github.com/microsoft/botbuilder-python/pull/1961

PR #1888, opened 2022-01-13
Fix Telemetry Logger attachments

### LINKED ISSUE #1880: An activity with attachments raises an exception in telemetry logger
## Version
4.14.1

## Describe the bug

Activity.attachments is a list. Therefore, telemetry logger raises an exception (see below) if trying to call `strip()` on attachements.

```
  File "/Users/ivo/Projects/alvao-chatbot/bot/venv/lib/python3.8/site-packages/botbuilder/core/telemetry_logger_middleware.py", line 220, in fill_send_event_properties
    if activity.attachments and activity.attachments.strip():
AttributeError: 'list' object has no attribute 'strip'
```

This line produces the bug: https://github.com/microsoft/botbuilder-python/blob/main/libraries/botbuilder-core/botbuilder/core/telemetry_logger_middleware.py#L220

## To Reproduce
1. Use TelemetryLoggerMiddleware
2. Create a card message: `MessageFactory.attachment(Attachment(content_type="application/vnd.microsoft.card.adaptive", content=card))`
3. Send this activity from bot to user
4. When this activity is being logged via TelemetryLoggerMiddleware, it raises an exception

## Expected behavior
There should be another check than `strip()`, possible `len() > 0`. Function `strip` returns a string without leading and trailing whitespaces. There is no sense to do that a) in a condition, b) on a list.

ramfattah:
Fixes #1880 An activity with attachments raises an exception in telemetry logger 

## Description

1) Removed `.strip()` method that was being called on a list as it was throwing `AttributeError: 'list' object has no attribute 'strip'` error.
2) Serialized the Attachment object to Json.


carlosscastro:
@axelsrz could you help drive this PR? Thanks :)

axelsrz (approved):

PR #2005, opened 2023-06-28
fix: Save a reference to created tasks, to avoid tasks disappearing

anio:
#minor

This change saves references for created tasks to avoid tasks disappearing. 

> Save a reference to the result of this function, to avoid a task disappearing mid-execution. The event loop only keeps weak references to tasks. A task that isn’t referenced elsewhere may get garbage collected at any time, even before it’s done.
>
> [python docs](https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task)

Please review the link: https://textual.textualize.io/blog/2023/02/11/the-heisenbug-lurking-in-your-async-code/


anio:
@microsoft-github-policy-service agree

tracyboehrer (approved):

PR #2022, opened 2023-10-04
Python 3.11 Compatibility

aacebo:
#minor

## Description
- remove deprecation warning in schema
- fix linting issues

## TODO
- another PR needs to be created to update the codebase and pipeline to support Python 3.11
> this would mean uncommenting [this](https://github.com/microsoft/botbuilder-python/blob/aacebo/py-3.11/pipelines/botbuilder-python-ci.yml#L11) and [this](https://github.com/microsoft/botbuilder-python/blob/aacebo/py-3.11/pipelines/botbuilder-python-ci.yml#L27) and updating the [pylint version](https://github.com/microsoft/botbuilder-python/blob/aacebo/py-3.11/pipelines/botbuilder-python-ci.yml#L59) to a version that supports 3.11. This will require you to update the `.pylintrc` and fix the codebase to conform to the new rules. 
- another PR needs to be opened to update dependencies so the packages can be imported into newer projects

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -11,7 +11,7 @@ ignore=CVS,build,botbuilder-schema,samples,django_tests,Generator,operations,ope
 
 # Add files or directories matching the regex patterns to the blacklist. The
 # regex matches against base names, not paths.
-ignore-patterns=setup.py,azure_bdist_wheel.py,**/tests/**/*.py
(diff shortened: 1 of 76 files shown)


tracyboehrer:
@aacebo These are failing the PyLint step:  pylint --rcfile=.pylintrc libraries

tracyboehrer (approved):

PR #2050, opened 2023-12-05
Dialogs: regex version bump

### LINKED ISSUE #2019: botbuidler support for  regex== 2022 and above
Description:

I'm currently working on building a chatbot using Azure Bot Builder SDK in conjunction with OpenAI. In my project, I'm relying on the OpenAIEmbedding class from the langchain package, which utilizes Tiktoken. However, I've run into an issue due to dependency conflicts with Tiktoken. Specifically, Tiktoken requires regex version 2022 or higher, while the Bot Builder package supports only up to regex version 2019.

Feature Request:

I kindly request adding support for Tiktoken's regex version 2022 or higher in the OpenAIEmbedding class within the langchain package. This update would resolve the dependency conflicts and enable smoother integration of OpenAI into projects using Azure Bot Builder SDK.

Additional Information:

Current Behavior: Currently, the OpenAIEmbedding class in langchain relies on Tiktoken, which necessitates a regex version that is not compatible with the Bot Builder SDK's regex version support.

Desired Behavior: The botbuilder  classes should be updated to support Tiktoken's dependency on regex version 2022 or higher t

Impact of the Feature:

This feature would benefit developers working on chatbot projects that use Azure Bot Builder SDK and OpenAI. It would eliminate dependency conflicts, allowing for a seamless integration experience.

tracyboehrer:
Fixes #2019

### DIFF
--- a/libraries/botbuilder-dialogs/setup.py
+++ b/libraries/botbuilder-dialogs/setup.py
@@ -5,7 +5,7 @@ import os
 from setuptools import setup
 
 REQUIRES = [
-    "regex<=2019.08.19",
+    "regex>=2022.1.18",
     "emoji==1.7.0",
     "recognizers-text-date-time>=1.0.2a1",
     "recognizers-text-number-with-unit>=1.0.2a1",


sandst1:
Nice, was just about to create an issue about this 👍 This should also fix a conflict with `semantic-kernel`, which requires regex `>=2023.6.3`

PR #2101, opened 2024-04-24
Fix update_activity in messages

NeeluGeorge:
closes: https://github.com/OfficeDev/Microsoft-Teams-Samples/issues/1226

## Description

Python sdk documentation says to pass replyToId as **updated_activity.id**, but this was not getting passed to the **reference** created in **update_activity**

https://learn.microsoft.com/en-us/microsoftteams/platform/bots/how-to/update-and-delete-bot-messages?tabs=python#update-cards

## Specific Changes
Added activity.id as the "activity_id" in reference

### DIFF
--- a/libraries/botbuilder-core/botbuilder/core/turn_context.py
+++ b/libraries/botbuilder-core/botbuilder/core/turn_context.py
@@ -231,7 +231,7 @@ class TurnContext:
         :return:
         """
         reference = TurnContext.get_conversation_reference(self.activity)
-
+        reference.activity_id = activity.id
         return await self._emit(
             self._on_update_activity,
             TurnContext.apply_conversation_reference(activity, reference),


tracyboehrer:
@NeeluGeorge What is the reason for this?  What is it solving?  'apply_conversation_reference' does this, except conditionally.

NeeluGeorge:
@tracyboehrer 'apply_conversation_reference' is using reference.activity_id
But TurnContext.get_conversation_reference(self.activity) called before this will map reference.activity_id with the original activity's id field.
Here we are passing **self.activity** in  TurnContext.get_conversation_reference(self.activity) ,  not the activity we pass as argument

So our updated activity.id is left unused.This throws invalid ID error

tracyboehrer:
@NeeluGeorge Thanks for that.  I'll dig a little.  I note that the Python get_conversation_reference is different from the DotNet and JS implementations.

tracyboehrer:
@NeeluGeorge reference.activity_id should already be activity.id.  That happens in get_conversation_reference.  But, Pythons get_conversation_reference is outdated, and that assignment should be conditional, which I will update.  So doing this could potentially change that logic.

That doc states the Activity passed to update_activity should have its id set to the incoming activity.reply_to_id.  Is there a value for that property?

NeeluGeorge:
> @NeeluGeorge reference.activity_id should be TurnContext.activity.id. That happens in get_conversation_reference. But, Pythons get_conversation_reference is outdated, and that assignment should be conditional, which I will update. So doing this could potentially change that logic.
> 
> That doc states the Activity passed to update_activity should have its id set to the incoming activity.reply_to_id. Is there a value for that property?

Yes @tracyboehrer 

Sample code for updation
`    async def post_updated_message(self):
        adaptive_card_attachment = CardFactory.adaptive_card(self.message.messageBody)
        updated_activity = MessageFactory.attachment(adaptive_card_attachment)
        updated_activity.id = self.turn_context.activity.reply_to_id
        await self.turn_context.update_activity(updated_activity)
  `
  **Issue in update activity**:  
  reference = TurnContext.get_conversation_reference(self.activity)
  
  Here self.activity is having a different **id**.
  Only our updated_activity has **id= 'reply_to_id'**
 But this is left unused while creating **reference**

NeeluGeorgeKB4:
@tracyboehrer 
Has this been fixed in any of the recent releases?

tracyboehrer:
In 4.15.1 there was a fix in CloudAdapter related to the id passed to update_activity.

PR #2144, opened 2024-07-26
Remove urllib3 upper bound version constraint

### LINKED ISSUE #2142: Support urllib3 version 2
**Is your feature request related to a problem? Please describe.**

My project uses urllib3 version 2, [released April 2023](https://urllib3.readthedocs.io/en/stable/changelog.html#id13). Now I’m trying to add botbuilder to it, but it’s not possible because `botbuilder-schema` pins it to version <2:

https://github.com/microsoft/botbuilder-python/blob/a2cccb32d1d23639f304288b6d7d970bb1bbd3ba/libraries/botbuilder-schema/setup.py#L9

Commit 80b35727f35b269e68b54234a2809b933de865a5 did this.

It’s not exactly clear why though, because urllib3 isn't used directly anywhere. 

**Describe the solution you'd like**

botbuilder-schema supports version 2 of urllib3. See [migration guide](https://urllib3.readthedocs.io/en/stable/v2-migration-guide.html).

**Describe alternatives you've considered**

Downgrading urllib3 but I have project code that uses v2-specific features.

**Additional context**
n/a

adamchainz:
Fixes #2142 

## Description

Remove the upper version bound constraint added in #2051. It’s unclear why it was added as @tracyboehrer left no notes. See [this mega blog post](https://iscinumpy.dev/post/bound-version-constraints/) for a long explanier on why libraries should avoid such constraints.

## Specific Changes

Per title

## Testing

Relying on CI.

### DIFF
--- a/libraries/botbuilder-schema/setup.py
+++ b/libraries/botbuilder-schema/setup.py
@@ -6,7 +6,10 @@ from setuptools import setup
 
 NAME = "botbuilder-schema"
 VERSION = os.environ["packageVersion"] if "packageVersion" in os.environ else "4.17.0"
-REQUIRES = ["msrest== 0.7.*", "urllib3<2.0.0"]
+REQUIRES = [
+    "msrest== 0.7.*",
+    "urllib3",
+]
 
 root = os.path.abspath(os.path.dirname(__file__))
 


tracyboehrer:
@adamchainz It's not uncommon for us to pin versions, often at least to major versions.  For packages that follow semver closely (not all do), that's typically safe.  Though not always.

adamchainz:
Sure, but you have to balance making your library safer against breakage versus blocking users from upgrading, which may be safer for them. I don’t want to downgrade urllib3 since it might break things in my project or make it less secure.

tracyboehrer:
@adamchainz I have no problems with the latest version, assuming it is verified to work.  Though I can't speak to some future 3.x version.

adamchainz:
...so will you merge and release this PR?

PR #2147, opened 2024-08-02
Fixed type hint in turn_context.py

### LINKED ISSUE #2146: Incorrect typehint in turn_context.py
## Describe the bug
send_activity function in botbuilder/core/turn_context.py has a type hint that implies that this function always returns ResourceResponse. That is not true, as it can also return None. This may lead other programmers, like me, to assume that there is never a need to check for None output here.

## Expected behavior
Type hint must be adjusted to include None

Nikita-Gz:
Previous type hint implied that send_activity in turn_context.py will always return ResourceResponse, meanwhile it can also return None

Fixes #2146

## Description
The type hint is fixed to include possible None return, otherwise it might make the developers assume that there is no need to check for None.

## Specific Changes
 - Changed ResourceResponse to Union[ResourceResponse, None] type hint

### DIFF
--- a/libraries/botbuilder-core/botbuilder/core/turn_context.py
+++ b/libraries/botbuilder-core/botbuilder/core/turn_context.py
@@ -158,7 +158,7 @@ class TurnContext:
         activity_or_text: Union[Activity, str],
         speak: str = None,
         input_hint: str = None,
-    ) -> ResourceResponse:
+    ) -> Union[ResourceResponse, None]:
         """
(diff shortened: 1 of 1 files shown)


tracyboehrer (approved):

Nikita-Gz:
"BotBuilder-Python-CI-PR-yaml" step seems to be waiting for something. Do you know what that step is about @tracyboehrer ? I can't seem to find its definition

PR #2154, opened 2024-08-12
Returning callback to get activity on proactive message send

### LINKED ISSUE #2153: Can't get activity ID when using ADAPTER.continue_conversation
## Version
botbuilder-core==4.16.1

## Describe the bug
When making proactive message with

```
activity_id = await ADAPTER.continue_conversation(
    reference,
    lambda turn_context: turn_context.send_activity(message),
    CONFIG.APP_ID,
)
print(activity_id)
```

it doesn't return anything, but I need to save activity id to update the message later

## To Reproduce
Use sample from https://github.com/microsoft/BotBuilder-Samples/blob/main/samples/python/16.proactive-messages/app.py#L87

When calling wait ADAPTER.continue_conversation save results to var like mentioned above.

## Expected behavior
I expect to have activity object to be able to save ID to modify this message later.

## Additional context
I can try to put a PR for this

afedotov-align:
Fixes #2153

## Description
Adding return statement to receive callback execution result to get activity id


Attels:
@microsoft-github-policy-service agree

Attels:
> @microsoft-github-policy-service agree

@microsoft-github-policy-service agree

PR #2213, opened 2025-03-12
Fix Middleware Chain to Allow Awaiting Final Logic Result

### LINKED ISSUE #2197: Middleware logic in turncontext emit is broken
## Version
4.16.2

## Describe the bug
The turncontext middlware chain is as follows:
https://github.com/microsoft/botbuilder-python/blob/a7f5d91398bf502128645ec5d698024c5aaeda3c/libraries/botbuilder-core/botbuilder/core/turn_context.py#L287-L304

Notice how the "logic" runs after all the middlware has run? This prevents any middlware from awaiting the result of the "next" middleware. This means that if you wanted to wait for the "response" from the last middleware (which runs the actual logic to send the activity), you can't.

This is in contrast with how the middleware for js is built:
https://github.com/microsoft/botbuilder-js/blob/3b8fcab21a0a5434706da8eace17117722ffd78b/libraries/botbuilder-core/src/turnContext.ts#L864.

Here, if you wanted to do:
```js
async function MyMiddleware(context, args, next) {
   console.log(args)
   const res = await next()
   console.log(res)
}
```
You can.

gandiddi:
Fixes #2197

## Description
Previously, the middleware chain executed each handler sequentially, but none could await the final logic result. This prevented middlewares from processing the response of the last middleware (which runs the actual logic).

This PR updates the `TurnContext` class in the `botbuilder-core` library to modify the middleware chain handling. The changes ensure that middleware can await the result of the "next" middleware, including the final logic.
This allows for more flexible and powerful middleware handling.

### 1. TurnContext Class:
- Modified the `_emit` method to ensure that each middleware can await the result of the next middleware in the chain, including the final logic.
- Updated the `_emit` method to return the result of the logic coroutine correctly.

### 2. Test Cases:
- Updated the `test_update_activity_should_apply_conversation_reference` test case to handle the `ResourceResponse` correctly and ensure the `update_activity` method returns the correct `ResourceResponse`.

### DIFF
--- a/libraries/botbuilder-core/botbuilder/core/turn_context.py
+++ b/libraries/botbuilder-core/botbuilder/core/turn_context.py
@@ -288,20 +288,15 @@ class TurnContext:
 
         async def emit_next(i: int):
             context = self
-            try:
-                if i < len(handlers):
+            if i < len(handlers):
+                try:
+                    return await handlers[i](context, arg, lambda: emit_next(i + 1))
+                except Exception as error:
+                    raise error
+            else:
+                return await logic
 
-                    async def next_handler():
-                        await emit_next(i + 1)
-
-                    await handlers[i](context, arg, next_handler)
-
-            except Exception as error:
-                raise error
-
-        await emit_next(0)
-        # logic does not use parentheses because it's a coroutine
-        return await logic
+        return await emit_next(0)
 
     async def send_trace_activity(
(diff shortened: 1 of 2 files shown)


rodrigobr-msft:
We will work on bringing this work into the Agents SDK for Python. Thank you for your contribution.a

PR #2215, opened 2025-03-14
fix: Method name is emit_event, not emit

### LINKED ISSUE #2214: [BUG] DialogEvent has no method emit when calling component dialog.
# Dialog Events don't work in ComponentDialogs

## Version
4.16.1, main branch at time of writing

## Describe the bug

  File ".../workspace/.venv/lib/python3.12/site-packages/botbuilder/dialogs/dialog.py", line 128, in on_dialog_event
    handled = await dialog_context.parent.emit(
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'DialogContext' object has no attribute 'emit'

## To Reproduce
1. Run emit_event on a component dialog

## Expected behavior
It doesn't throw an error

alexrecuenco:
Fixes #2214

## Description

The function name is emit_event, not emit

## Specific Changes
  - Modified the name
  

## Testing
I would need guidance on what test to modify that currently has emit_event in the codebase.

### DIFF
--- a/libraries/botbuilder-dialogs/botbuilder/dialogs/dialog.py
+++ b/libraries/botbuilder-dialogs/botbuilder/dialogs/dialog.py
@@ -125,7 +125,7 @@ class Dialog(ABC):
 
         # Bubble as needed
         if (not handled) and dialog_event.bubble and dialog_context.parent:
-            handled = await dialog_context.parent.emit(
+            handled = await dialog_context.parent.emit_event(
                 dialog_event.name, dialog_event.value, True, False
             )
 


axelsrz:
@alexrecuenco were you able to hit this bug in actual usage?

From an initial investigation, it looks like this was code used as preparation for another set of features that are not currently in this SDK

alexrecuenco:
> @alexrecuenco were you able to hit this bug in actual usage?
> 
> 
> 
> From an initial investigation, it looks like this was code used as preparation for another set of features that are not currently in this SDK

Yes. I emit events, and then those events get captured by a parent component to do some resource refresh handling


My hotfix is to monkey-patch emit to be equal to emit event.

The feature exists in the botbuilder-js, and here it is working perfectly well once you do that.

alexrecuenco:
> Yes. I emit events, and then those events get captured by a parent component to do some resource refresh handling
> 
> 

In particular, this is doing some life cycle generic component, that I am hoping to open source soon.

Any resource that is generated through a dialog with a user is encapsulated.

In particular, the main use is something akin to an authorization middleware within a bot. So you have an AuthorizedComponent that generates the sufficient context, and this component encapsulates token authorization. 

The internal dialog then has in the turn state the reference to the required resource/authorization, without having to manually handle authorization in each dialog.

And for long running tasks, internal resources can extend that token by calling refresh(dialog_context), that in turns emits an event and is captured by the nearest matching AuthorizedComponent that refreshes the token/resource.

axelsrz (approved):

alexrecuenco:
You approved it, but does it need anything else to he merged? 

I see it is still not executing the tests, it says that it needs maintainer permissions.

PR #2219, opened 2025-03-23
feat(dialogs): Incorporate DialogDependencies

### LINKED ISSUE #2218: Add DialogDependencies, working towards feature parity with botbuilder-js and botbuilder-dotnet
### Use this [query](https://github.com/Microsoft/botbuilder-python/issues?q=is%3Aissue+is%3Aopen++label%3Afeature-request+) to search for the most popular feature requests.

**Is your feature request related to a problem? Please describe.**
It is extremely useful when building dialog frameworks, to prevent having to juggle the dialogs and then manually having to add them. Forgetting dependencies is a constant issue. 

Or even worse, hooking to a dialog that you thought was related, but it turns out is just another dialog with the same name, and your dialog you just forgot to add it

**Describe the solution you'd like**

A Python Protocol called DialogDependencies that when implemented on a dialog, it allows you to define dependencies of dialogs and allows for them to be added automatically when you call `dialog_set.add(main_dialog)`

**Describe alternatives you've considered**
NA

**Additional context**

Dialog dependencies is a feature in

- [botbuilder-js](https://github.com/microsoft/botbuilder-js/blob/87d0ad1a889a38396e761f10614c78a4526590d1/libraries/botbuilder-dialogs/src/dialogSet.ts#L169),
- and in [botbuilder-dotnet](https://github.com/microsoft/botbuilder-dotnet/blob/bc17a5db2c717fec0d4c109d884bea88d29f75cb/libraries/Microsoft.Bot.Builder.Dialogs/DialogSet.cs#L147).

alexrecuenco:
Fixes #2218

## Description

Dialog dependencies is a feature in

- [botbuilder-js](https://github.com/microsoft/botbuilder-js/blob/87d0ad1a889a38396e761f10614c78a4526590d1/libraries/botbuilder-dialogs/src/dialogSet.ts#L169),
- and in [botbuilder-dotnet](https://github.com/microsoft/botbuilder-dotnet/blob/bc17a5db2c717fec0d4c109d884bea88d29f75cb/libraries/Microsoft.Bot.Builder.Dialogs/DialogSet.cs#L147).

To maintain feature parity, and because this functionality is extremely useful this PR brings it to `botbuilder-python`.

## Specific Changes

Tecnically in the `botbuilder-js` and dotnet implementations dialogs that are found to conflict get stored with a different ID, but I am sure that would likely be a breaking change (and IMO a very counter-intuitive way of working).

If you are adding dialog dependencies and they conflict, the logic as it was would be retained.

Note that, although by making it a `@runtime_checkable` Protocol with DialogDependencies one can check `isinstance(object, DialogDependencies)`, the recommendation from the python docs is:

"check against a runtime-checkable protocol can be surprisingly slow" — [python typing](https://docs.python.org/3/library/typing.html#typing.Protocol)

For that reason we use `hasattr` and `callable` instead.

Also, note that protocol is avaialble from python 3.8 onwards, which is already the minimum version for this library.

The purpose of using Iterable instead of List is just to keep it generic. Since python has the concepts of `tuple`, `List`, etc and to keep it as a structural type.

## Testing

- [x] Tests with coverage is added
- [x] pylint was run against the code, (although note that the pre-commit used uses a dependency that doesn't work in python 3.12, updating to `repo: https://github.com/pylint-dev/pylint` would be nice
- [x] black was run against the code (note that some libraries are actually not properly formatted according to black!)

### DIFF
--- a/libraries/botbuilder-dialogs/botbuilder/dialogs/dialog_set.py
+++ b/libraries/botbuilder-dialogs/botbuilder/dialogs/dialog_set.py
@@ -2,7 +2,7 @@
 # Licensed under the MIT License.
 import inspect
 from hashlib import sha256
-from typing import Dict
+from typing import Dict, TYPE_CHECKING
 
 from botbuilder.core import (
     NullTelemetryClient,
@@ -13,6 +13,9 @@ from botbuilder.core import (
 )
 from .dialog import Dialog
 from .dialog_state import DialogState
+
+if TYPE_CHECKING:
+    from .dialog_context import DialogContext
 
 
 class DialogSet:
@@ -92,6 +95,8 @@ class DialogSet:
             )
 
         if dialog.id in self._dialogs:
+            if self._dialogs[dialog.id] == dialog:
+                return self
             raise TypeError(
                 "DialogSet.add(): A dialog with an id of '%s' already added."
                 % dialog.id
@@ -99,6 +104,11 @@ class DialogSet:
 
         # dialog.telemetry_client = this._telemetry_client;
         self._dialogs[dialog.id] = dialog
+
+        # Automatically add any child dependencies the dialog might have, see DialogDependencies.
+        if hasattr(dialog, "get_dependencies") and callable(dialog.get_dependencies):
+            for child in dialog.get_dependencies():
+                self.add(child)
 
         return self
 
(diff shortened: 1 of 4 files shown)


libraries/botbuilder-dialogs/botbuilder/dialogs/dialog_set.py
```diff
@@ -14,6 +14,9 @@
 from .dialog import Dialog
 from .dialog_state import DialogState
 
+if TYPE_CHECKING:
+    from .dialog_context import DialogContext
```
alexrecuenco, line 18: This line is necessary for the `-> "DialogContext"` used below, otherwise intellisense has a hard time understanding that the object returned is a `DialogContext`. It has no runtime impact because of `TYPE_CHECKING` being `False` at runtime

PR #2220, opened 2025-03-25
Show typing middleware bugfix

### LINKED ISSUE #1980: ShowTypingMiddleware - Infinite loop of typing activity when Adapter on_error() method is called.
## Version
I am using the following packages
```
botbuilder-schema==4.13.0
botbuilder-core==4.13.0
botbuilder-dialogs==4.13.0
```

## Describe the bug
ShowTypingMiddleware goes into an infinite loop of sending typing activity to user when on_error() method of the BotFrameworkAdapter is called. It doesn't stop till the application is restarted. This is easily reproduced in BotBuilder-Samples tested with BotFramework Emulator.

## To Reproduce
1. add `ADAPTER.use(ShowTypingMiddleware(delay=0.5, period=1.0))` to app.py of [13, Core Bot Python Sample](https://github.com/microsoft/BotBuilder-Samples/tree/main/samples/python/13.core-bot) 
2. Write some code that takes calls the adapter's on_error() method i.e. `print(ThisIsABug)` in destination step in dialogs/booking_dialog.py
3. Start the app and test the dialog from BotFramework Emulator. 
4. The app sends infinite typing activity to the emulator when the error is encountered in code.

## Expected behavior
ShowTypingMiddleware should stop sending typing activity if on_error() method is called.

## Screenshots
Screenshot from 2022-11-01 14-38-03

## Additional context
NA

alexrecuenco:
Fixes #1980

## Description

The typing middleware creates a long running task that is not cleared 

## Specific Changes

  - The first commit shows a reproducible test of this bug
  - The second commit fixes the bug

## Testing

The tests are added on the first commit.

### DIFF
--- a/libraries/botbuilder-core/botbuilder/core/adapters/test_adapter.py
+++ b/libraries/botbuilder-core/botbuilder/core/adapters/test_adapter.py
@@ -699,6 +699,7 @@ class TestFlow:
                     reply = adapter.activity_buffer.pop(0)
                     raise RuntimeError(
                         f"TestAdapter.assert_no_reply(): '{reply.text}' is responded when waiting for no reply."
+                        f"(Activity type: {reply.type})"
                     )
 
                 await asyncio.sleep(0.05)
--- a/libraries/botbuilder-core/botbuilder/core/show_typing_middleware.py
+++ b/libraries/botbuilder-core/botbuilder/core/show_typing_middleware.py
@@ -93,12 +93,10 @@ class ShowTypingMiddleware(Middleware):
         ):
             start_interval(context, self._delay, self._period)
 
-        # call the bot logic
-        result = await logic()
-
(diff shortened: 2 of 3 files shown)


libraries/botbuilder-core/botbuilder/core/adapters/test_adapter.py
```diff
@@ -699,6 +699,7 @@ async def wait_for_activity():
                     reply = adapter.activity_buffer.pop(0)
                     raise RuntimeError(
                         f"TestAdapter.assert_no_reply(): '{reply.text}' is responded when waiting for no reply."
+                        f"(Activity type: {reply.type})"
```
alexrecuenco, line 702: I am adding this extra text, because otherwise you get a "None is responded" without further feedback in the runtime assertion error.

